PR #7, opened 2016-01-21
Add language setting for tesseract

the01:
With the new setting you can specify the language tesseract should use for the ocr (might improve results)

additional languages can be found at https://github.com/tesseract-ocr/tessdata

### DIFF
--- a/src/documents/management/commands/document_consumer.py
+++ b/src/documents/management/commands/document_consumer.py
@@ -37,6 +37,7 @@ class Command(BaseCommand):
     CONSUME = settings.CONSUMPTION_DIR
 
     OCR = pyocr.get_available_tools()[0]
+    OCR_LANG = settings.TESSERACT_LANGUAGE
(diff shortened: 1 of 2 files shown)


danielquinn:
My system (Gentoo) only comes with English by default, so I'll have to figure out how to set it up to "train" Tesseract based on those language files, but I agree, this is an important issue.  I've got a bunch of Dutch documents that I'd like to have scanned with a proper language file for example.

The tricky part is to how to tell Tesseract *in a non-interactive way* that I want to OCR this document with *language X* instead of the default *language Y*.  Given that the whole idea behind this is to push 1 button and forget about the document completely, I don't want to have to introduce some sort of manual switch in the app.

Ideally, some form of detection would be ideal, or barring that, a means of telling the consumer to expect *language X* from directory *X* and *language Y* from directory *Y*.  The latter is more limiting though, since my scanner only supports 2 target directories, and I'm willing to bet that most only support 1.

PR #26, opened 2016-02-12
Document cloning from public URL rather than ssh

wttw:
Trivial docs change to point to public https URL on github.


danielquinn:
Good idea!

PR #31, opened 2016-02-14
Enable parallel OCR processing

pitkley:
At the moment, every page in a PDF will be processed one by one using tesseract. Since the processing of a single page is independent from every other page, one can make use of multi-core machines.

This PR introduces a multiprocessing pool to process multiple pages simultaneously. The amount of threads to use can be specified in the environment variable `PAPERLESS_OCR_THREADS`. This will default to the number of cores/hyperthreads Python detects for your system.

I have not written documentation for this feature yet, I would greatly appreciate a pointer in which file/section this information should be included!

Furthermore, this change might be incompatible to the `feature/images-as-docs` branch, I hadn't had the chance to check yet.

PR #32, opened 2016-02-14
Detect language only on first page of PDF

pitkley:
To detect the language currently the entire document gets processed. If a different language has been detected than the default one, the entire document will be processed again for the new language.

This PR analyzes the first page for its language and either processes the remaining pages with the default language if it didn't differ, or processes all pages for the new guessed language.

The amount of processed pages comes down from the worst case `2n` to worst case `n+1`.

One thing I worry about: What if the language is not apparent on the first page, especially if it is some kind of titlepage? While I would like to reduce the amount of pages to process as far as possible, scanning just the first page might be the wrong way to do it. Maybe pick one at random? I don't know, I would love some suggestions.

(BTW: This PR is compatible and doesn't conflict with `feature/paralellism` in #31.)

### DIFF
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -9,6 +9,7 @@ import random
 import re
 import subprocess
 
+import math
 import pyocr
 
 from PIL import Image
@@ -152,9 +153,14 @@ class Consumer(Renderable):
         simple language detection trial & error.
         """
 
+        if not pngs:
+            raise OCRError
+
         self._render("  OCRing the document", 2)
 
-        raw_text = self._ocr(pngs, self.DEFAULT_OCR_LANGUAGE)
+        # Since the division gets rounded down by int, this calculation works for every edge-case, i.e. 1
+        middle = int(len(pngs) / 2)
+        raw_text = self._ocr([pngs[middle]], self.DEFAULT_OCR_LANGUAGE)
 
         guessed_language = self._guess_language(raw_text)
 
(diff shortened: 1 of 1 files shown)


danielquinn:
Actually, I'd opt for the middle page, so something like `pages[int(len(pages)/2)]`.  That way you're most likely to avoid the title page & empty last page thing.  What do you think?

pitkley:
Yes, something along those lines might be a good idea. I'll give the middle page a try!

pitkley:
I have updated the logic to use the middle page (which will be the first page for documents with one or two pages). I've tried it with documents of lengths 1, 2, 3 and 8 for both the case that the default language matched and the one where it didn't and got the expected results.

src/documents/consumer.py
```diff
@@ -191,6 +200,9 @@ def _ocr(self, pngs, lang):
         Performs a single OCR attempt.
         """
 
+        if not pngs:
```
danielquinn: I'm a little worried about this bit.  If `pngs` is empty, then we should be blowing up somewhere, not silently returning `""`, no?  Did you run into a case where this happened?  Perhaps we can fix the cause rather than handle a case where this would happen?

pitkley: I'm slicing the PNG-list in `_get_ocr` into `headlist - middle - taillist`. For documents of lengths 3 and up, both the `headlist` and the `taillist` will be populated, for length 2 the `headlist` will be empty and for length 1 both the `headlist` and the `taillist` will be empty.

I moved the check into here to not clutter the `_get_ocr` implementation with checks for empty lists all over the place but rather ignore it in a central place.

What we could and probably should do though is add a check into the top of `_get_ocr` and fail if the initial list we get is empty. What do you think?

danielquinn: I see what's going on now, thanks.

As for adding a check at the top of `_get_ocr()`, I suppose it'd be good accounting, even though that *should* never happen ("should", famous last words).  Alright, if you add a check at the top as you suggest, I'll merge this bit happily.

pitkley:
I've added the additional check, from my side it would be clear for merge, no late fixes this time around (hopefully...).

danielquinn:
Looks good to me, thanks!

PR #35, opened 2016-02-14
Fix matching if user supplied an empty value

pitkley:
In my tests, every single tag that had no value set for "match" was automatically applied to every new document, probably because of the change in fc5d89c.

I don't know if this is how it is supposed to work now, but I would like to be able to have tags not matching anything. This change checks if the user supplied an empty value and if so doesn't match the tag.


danielquinn:
It never occurred to me that someone would want to have a tag with an value, but this makes sense.

PR #38, opened 2016-02-15
Update consumer imports to be relative (to make IDE inspections happy)

jat255:
Little change to make imports relative (makes my IDE happy)

### DIFF
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -17,8 +17,8 @@ from django.conf import settings
 from django.utils import timezone
 from django.template.defaultfilters import slugify
 
-from logger.models import Log
-from paperless.db import GnuPG
+from ..logger.models import Log
+from ..paperless.db import GnuPG
 
 from .models import Sender, Tag, Document
 from .languages import ISO639


danielquinn:
I'm not sure how I feel about this.  I think this may be a problem with your IDE rather than the code.

Obviously, relative imports are better than concrete ones, but in a Django project you typically only use relative imports from within the app (in this case `documents`).  Django apps are usually seen as separate entities on their own, so you don't do relative imports between them.

If your IDE is barking, make sure that it knows that `src` is in the Python path.  In my case (PyCharm) I just right-click on the `src` directory in the side menu and toggle `Mark As > Sources Root`.

I'm going to leave this open though 'cause I want to see what other people think about this.  I'm also going to check with some Django devs in meatspace to see if I've been violating convention all these years.

jat255:
Oh, I am using PyCharm, and setting `src` as the root fixed the problem on my end. If you end up wanting to close this, that's fine with me.

danielquinn:
So I checked with another dev on this one and the verdict is in: as Django apps are meant to be self-contained, *relative links only within apps*.  Thanks for making me reexamine my own coding practises though! :-)

PR #40, opened 2016-02-15
Enabling django-suit theme

jat255:
Please feel free to reject this if it is unwelcome. As I've been playing around trying to learn Django a little bit, I stumbled upon `django-suit`, which provides an easy way to switch to a more modern looking admin interface. I enabled it locally and it doesn't seem to have had any negative impacts, and seems to be more responsive in terms of navigating around the database.

Feel free to checkout my branch to test it out!

Main interface screen:
image

Documents page:
image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 Django==1.9
 django-extensions==1.6.1
+django-suit==0.2.16
 filemagic==1.6
 langdetect==1.0.5
 Pillow==3.0.0
--- a/src/manage.py
+++ b/src/manage.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import os
 import sys
+from django.conf.global_settings import TEMPLATE_CONTEXT_PROCESSORS as TCP
 
 if __name__ == "__main__":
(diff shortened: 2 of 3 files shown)


danielquinn:
Yeah the Django-suit theme is pretty sexy isn't it?  The problem is that I'm not sure if the licenses are compatible or not.  Paperless is GPL3 and django-suit has a special license that limits use under some circumstances.  Rolling them together in the same app may be legally problematic.

However, I don't think it's worth worrying about, since the Django admin was never meant to be the primary interface.  I'm currently working on setting up a standard REST API so I can then write a nice user-friendly, (hopefully more attractive) interface on top.  Theoretically we could even make things skinnable.

danielquinn:
So yeah, nice idea, but for the reasons above, I'm going to have to reject this PR (sorry).

jat255:
Not a problem! I figured there might be something like that. I think I'm safe to keep using it locally, though? There are also some other nice themes that are available under some more open licenses, but if you're going to writing a new GUI anyway, it's probably not necessary to incorporate anything.

danielquinn:
Nope, there's no harm at all in you continuing to use suit locally, go nuts!

PR #50, opened 2016-02-19
Some small tweaks to the Docker setup and documentation

tikitu:
These are leftovers from #39 which I realised I could better do myself, rather than holding up merging that PR. If any are contentious I'm happy to spin them out into separate PRs or remove them entirely.

* [x] Simply export instructions by adding a volume `/export` to `docker-compose.yml.example`
* [x] Since users will modify `docker-compose.env`, it also should follow the "copy `x.example` to `x` and make local modifications" pattern


pitkley:
I have added some minor nit-picks to the diff :)

tikitu:
Thanks, good stuff! I got this far while juggling a baby over breakfast, and will have to park it until the weekend to fold that feedback in.

danielquinn:
I agree with @tikitu that we shouldn't reuse the consumption directory as a place to put exports, but I'll let you two settle on *how* that is best defined since you're the Docker people :-)  When you're both happy with what's in this PR, please post a :+1: each and I'll know we can merge.

tikitu:
I think I dealt with all of @pitkley's feedback -- thanks for the careful attention! :+1: from me to merge.

.gitignore
```diff
@@ -69,6 +69,7 @@ db.sqlite3
 virtualenv
 .vagrant
 docker-compose.yml
+docker-compose.env
```
danielquinn: if `docker-compose.env` is being removed from git, should it not be replaced with a `docker-compose.env.example`?

danielquinn:
Everything looks good to me.  Do we have consensus gentlemen?

pitkley:
Yes, looks good to me :+1:

PR #62, opened 2016-02-23
Compensate for case and format of jpg vs. jpeg

danielquinn:
This fixes an ugly bug that occurred when the consumer indexed a jpeg file that was suffixed with either `.jpeg` or even `.JPG`.  The result was that the file was indeed consumed, given an id, encrypted and stored in the media directory, but it was unavailable via the web interface because of naming.

This merge fixes all of that (hopefully), but it doesn't provide a means of getting your mis-named files out of paperless.  For that, you need to do two things:

1. Rename the files in the media directory.  This is pretty simple, and if you have the standard `rename` utility on your system, even easier.  Basically you need to change all files in the media directory that contain `jpeg` to `jpg`:

        $ cd /path/to/media/documents/
        $ rename jpeg jpg *

2. Next, you need to tell the database that the file type has changed.  To do that, you can use the Django shell since that's easiest:

        $ cd /path/to/paperless/src/ (the location of manage.py)
        $ ./manage.py shell
        (now that you're in the Django shell)
        from documents.models import Document
        Document.objects.filter(file_type="jpeg").update(file_type="jpg")

    ...and that's it.  If you're uncomfortable with the Django shell, you can always go with sqlite:

        $ cd /path/to/paperless/src/ (the location of manage.py)
        $ ./manage.py dbshell
        UPDATE documents_document SET file_type = 'jpg' WHERE file_type = 'jpeg';

### DIFF
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -264,6 +264,12 @@ class Consumer(object):
                     Tag.objects.get_or_create(slug=t, defaults={"name": t})[0])
             return tuple(r)
 
+        def get_suffix(suffix):
+            suffix = suffix.lower()
+            if suffix == "jpeg":
+                return "jpg"
+            return suffix
+
         # First attempt: "<sender> - <title> - <tags>.<suffix>"
         m = re.match(self.REGEX_SENDER_TITLE_TAGS, parseable)
         if m:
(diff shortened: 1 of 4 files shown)


zedster:
That worked for me thank you,
although on ubuntu "rename JPG jpg *" didn't work as rename expected a regular expression.
So in-case it helps anyone else: i used "mmv \*.JPG.gpg \#1.jpg.gpg"

danielquinn:
Ah right, Ubuntu has `prename` in symlinked to `/usr/bin/rename` by default because *reasons*.  I believe the traditional rename utility is still available on Ubuntu systems, but under some other name.

PR #65, opened 2016-03-02
Update env-var in Dockerfile, fix volume names

pitkley:
That might be needed for Vagrant too, although I haven't checked it out.


danielquinn:
Thanks a lot for this.  I'll take a look at the Vagrant one today.

PR #73, opened 2016-03-06
Update Dockerfile to match latest version

pitkley:


danielquinn:
I don't understand, didn't these directories come in as part of the git pull?  I put `.keep` files in each of those so they'd be preserved in git.

pitkley:
Initially I only copied the `src` directory, I have expanded it to the `data` and `media` directories. This should future-proof the Dockerfile as long as those two base-directories don't change (if they do, bigger modifications would have to be done anyway).

pitkley:
Thanks, super quick :+1:

Issue #20, opened 2016-02-11
Deskew/Despeckle


rileytg:
Often times when scanning documents (especially in bulk) you run into situations where the document is slightly skewed and/or has speckles that throw off OCR.

![skew](http://www.verypdf.com/wordpress/wp-content/uploads/2015/04/image_thumb6.png)
![speckle](http://www.fmwconcepts.com/misc_tests/cc_noise_remove/jx8MPc9_kuw2_lat10.png)

rileytg:
Just opening this issue for thoughts on how to automatically process for this. There are a number of commercial products but I haven't found anything OSS that I've gotten to work well

danielquinn:
I hear that [unpaper](https://www.flameeyes.eu/projects/unpaper) is the go-to tool for this sort of thing, but integrating it would have to be done on the command-line level (like we're already doing for Tesseract).  I'm not opposed to this, but given that all of my scans so far have been really good quality, I'm not prioritising it.

However, if someone wants to write a PR to make this work, I'd probably merge it :-)

strages:
This would be incredible!

pitkley:
Just to keep you guys in the loop: I have a working version of Paperless with `unpaper` which you can find over at [`feature/unpaper`](https://github.com/pitkley/paperless/tree/feature/unpaper).
This PR is currently blocked by #34, but as soon as that is resolved I will open a PR for `unpaper`.

I did some limited testing (mainly because I don't have any bad scan samples at hand), but from what I have seen, the OCR results at least didn't get worse for decent scans.

If somebody (a) either has a bad scan for me to test with or (b) can test it for themselves, some feedback would be great!

Flameeyes:
If there is any help I can lend with unpaper please don't hesitate to tag me :)

danielquinn:
@Flameeyes as @pitkley suggested, you could post a link to a low-quality scan you'd like to see work, or even try out his fork and test some stuff yourself :-)

Flameeyes:
Sorry I should have been more clearer, I meant that as a current unpaper
maintainer :)

Although I'd be happy to try this out once I have some spare time!

On Wed, Feb 17, 2016 at 8:36 AM Daniel Quinn <notifications@github.com>
wrote:

> @Flameeyes <https://github.com/Flameeyes> as @pitkley
> <https://github.com/pitkley> suggested, you could post a link to a
> low-quality scan you'd like to see work, or even try out his fork and test
> some stuff yourself :-)
>
[...]

PR #74, opened 2016-03-06
Add unpaper

pitkley:
This is in regards to issue #20 and adds `unpaper` as another pre-processing step.

### DIFF
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -189,6 +189,9 @@ GNUPG_HOME = os.getenv("HOME", "/tmp")
 # Convert is part of the ImageMagick package
 CONVERT_BINARY = os.getenv("PAPERLESS_CONVERT_BINARY")
 
+# Unpaper
+UNPAPER_BINARY = os.getenv("PAPERLESS_UNPAPER_BINARY", "unpaper")
+
 # This will be created if it doesn't exist
 SCRATCH_DIR = os.getenv("PAPERLESS_SCRATCH_DIR", "/tmp/paperless")
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ MAINTAINER Pit Kleyersburg <pitkley@googlemail.com>
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         sudo \
-        tesseract-ocr tesseract-ocr-eng imagemagick ghostscript \
+        tesseract-ocr tesseract-ocr-eng imagemagick ghostscript unpaper \
     && rm -rf /var/lib/apt/lists/*
 
 # Install python dependencies
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -10,11 +10,13 @@ should work) that has the following software installed on it:
 * `GNU Privacy Guard`_
 * `Tesseract`_
 * `Imagemagick`_
+* `unpaper`_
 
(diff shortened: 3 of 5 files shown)


pitkley:
I have added `unpaper` to the `vagrant-provision` script, but I have not tested it. I don't know if unpaper is available in the `ubuntu/trusty64` box (although it should).

danielquinn:
I'm in the middle of a Very Big Blog Post right now, so I can't test this right away, but at first glance it looks good.  A couple questions though:

You've added `use_unpaper()` as a separate function rather than a method of Consumer.  Is that necessary?  I'd rather not have it floating around unattached unless there's a compelling reason for it.  Similarly, you've removed my comment about `image_to_string()` having to be a separate function, but maybe you can tell me why it can't be a class method?

I like things to be tidy and wrapped up into parent classes is all, so when that doesn't happen, I like there to be an explanation.

As for the vagrant bit, thanks for adding it.  I'll spin it up and give it a try before merging.

src/paperless/settings.py
```diff
@@ -248,6 +251,9 @@ def deprecated(before, after):
         deprecated("PAPERLESS_CONVERT", "PAPERLESS_CONVERT_BINARY")
         CONVERT_BINARY = os.getenv("PAPERLESS_CONVERT", CONVERT_BINARY)
 
+if not UNPAPER_BINARY:
```
danielquinn: Is this necessary?  Given that the name is unchanged, why not set [line 193](https://github.com/danielquinn/paperless/pull/74/files#diff-21758e26d686720f35a56d797f195c76R193) to this instead?

```python
UNPAPER_BINARY = os.getenv("PAPERLESS_UNPAPER_BINARY", "unpaper")
```

pitkley: It's not, I just thought that was the style you wanted to go for. I'll change it to use the default-param!

danielquinn: Ah ok.  No worries.  This whole block here at the bottom is just to handle older installations that use the deprecated environment variables.  Since `UNPAPER_BINARY` is new, it's not necessary.

pitkley:
> but maybe you can tell me why it can't be a class method?

Sorry, forgot to mention. For the parallelization to work, Python needs to serialize the class. The added `logger` field is not serializable though, causing it to fail with the error you experienced. This is true for both `image_to_string` and `run_unpaper`. 

The alternative would be to add a `__getstate__` method which removes the logger-reference so that serialization doesn't fail. AFAIK that is considered a fine way to do things and not some hacky work-around.

If you would rather have both methods to be part of the class, I can open a separate PR to add the serialization-fix and then adapt this one too.

danielquinn:
No, now that you've explained why it has to be this way, I think it's fine the way it is.  I may add a comment to that section later though, roughly equivalent to what you've just said here -- just to prevent me from meddling in the future :-)

pitkley:
Okay, cool. I have updated the environment-variable to use a default!

danielquinn:
Cool beans.  Alright after I'm finished this blog post, I'll give your PR a shot.  Should be a few hours though.

danielquinn:
I ran a few tests and the output is objectively better!  Thanks for this.

pitkley:
> [...] and the output is objectively better!

Awesome. I have not thoroughly compared results before and after, so that is great to hear!

Issue #51, opened 2016-02-19
Data loss when exporting documents with duplicate titles


tikitu:
The export script writes documents into a directory without checking for name collisions. This means that if I upload two documents with the same title, they will produce distinct entries in the database, but the export will clobber one of them.

I'm happy to tackle fixing this (I have quite some more experience with python and django than with docker :-) ), but wanted to get your pre-approval for the approach I would take. I thought to add a timestamp prefix to the export, so that files come out looking like `2016-02-13T23:15:02Z - Title - tag,tag.pdf`. It's not 100% watertight, but it seems pretty much good enough to me. It also implies the consumer has to recognise this format -- that has one nice side effect, namely that exporting docs and re-importing them will also preserve the timestamps at which they were *originally* imported.

If you prefer a different approach, I'm happy to take it on if you can give me a sketch, or to brainstorm alternative ideas here. I actually need something like this, because one of my use cases for paperless is storing documents that come in at regular intervals and will always get the same name (e.g. the electricity bill).

* [ ] Are filenames with `.isoformat()` datetimes scp-compatible? (There is some concern about the `:` character.)

danielquinn:
Ouch.  I hope you didn't lose any documents thanks to this bug!  Thanks for noticing and for offering to fix it, since I'm really busy for the next couple days.

Regarding your proposed fix, the timestamp is actually being exported appropriately by [re-writing the creation time on the document](https://github.com/danielquinn/paperless/blob/master/src/documents/management/commands/document_exporter.py#L52), so including it in the name is sort of redundant.  Additionally, messing with the name of the file on output means that importing it later would involve changing the name of the document in the database.  That is to say, a document with the name `2016-02-13T23:15:02Z - Title - tag,tag.pdf` would have a sender of `2016-02-13T23:15:02Z`.

I think I'd rather go the route of specifically changing the title to something that doesn't conflict.  So if `Sender - Title - tag,tag,tag.pdf` already exists, we could change it to `Sender - Title 2 - tag,tag,tag.pdf`.

Perhaps a better solution, albeit a more complicated one, would be to extend the naming regex to account for conflict addendums like ` 2`, so the output file would look like `Sender - Title - tag,tag,tag 2.pdf`.  This has the advantage of exported files being imported exactly like you'd expect, but it would mean fiddling with the name matcher on import.  Thankfully, this is one of the few places for which we have [tests](https://github.com/danielquinn/paperless/blob/master/src/documents/tests/test_consumer.py).

[...]

PR #79, opened 2016-03-07
Refactor file info extraction

tikitu:
Here's the useful bit of my (slow) movement towards fixing #51, before that became unnecessary.

There are some failing tests: they're edge cases that I turned up in the regex handling. I think everything should work as it used to, but I haven't given it *very* close attention: it turns out my lifestyle (small children; day job) is not really compatible with the pace of development you're keeping up here.

### DIFF
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -24,7 +24,7 @@ from pyocr.tesseract import TesseractError
 
 from paperless.db import GnuPG
 
-from .models import Correspondent, Tag, Document, Log
+from .models import Correspondent, Tag, Document, Log, FileInfo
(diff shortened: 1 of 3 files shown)


danielquinn:
This is a big PR, so I'll need some time to look over it.  Let me get back to you in a couple days?

danielquinn:
I'm sorry this is taking so long, but I have been poking at it locally.  Don't think I've forgotten!

danielquinn:
Finally merged!  Sorry it took so long, but I wanted to (a) account for the test failures, and (b) add support for including the date in the file name.

tikitu:
Hey, great!

PR #82, opened 2016-03-08
Fix Django Project URL - Previous link dead

bodegard:


danielquinn:
Oops.  My bad.  Thanks for noticing!

PR #86, opened 2016-03-15
Add brew install libmagic to apple-tastic requirements.

stevenvandervalk:
`./manage.py migrate` ran into `raise ImportError('Unable to find magic library')` on latest Apple tastic.

Perhaps I missed something else.


danielquinn:
Paperless makes use of libmagic for file type detection, so yes, good call on that one!

PR #90, opened 2016-03-19
Add directions for automatic starting at bootup on Ubuntu 14.04

stringlytyped:
I recently installed Paperless on a Ubuntu Server LTS 14.04 machine and thought I would document the steps needed to get it to run automatically at startup, in case it helps anyone else.

I figure that the _Making Things a Little more Permanent_ section of the documentation probably should be expanded further for systemd and Vagrant as time goes on.


danielquinn:
I'm always happy to accept more documentation -- especially to support systems I don't personally use.  Thanks!

PR #93, opened 2016-03-19
Remove PAPERLESS_OCR_LANGUAGES from docker-compose.yml

djmaze:
It is defined in `docker-compose.env` already.

### DIFF
--- a/docker-compose.yml.example
+++ b/docker-compose.yml.example
@@ -11,8 +11,6 @@ services:
             - data:/usr/src/paperless/data
             - media:/usr/src/paperless/media
         env_file: docker-compose.env
-        environment:
-            - PAPERLESS_OCR_LANGUAGES=
         command: ["runserver", "0.0.0.0:8000"]
 
     consumer:


danielquinn:
I'm going to run this by @pitkley before I just merge it blindly.  You cool with this Pit?

pitkley:
[This is intentional](https://github.com/danielquinn/paperless/pull/39#issuecomment-185695770) and should not be removed:

> > Can we not remove this line and port that bit to the environment file?
>
>The reason the line is there is so that the webserver that doesn't do any text recognition doesn't have to install unnecessary languages the user might have set in the env-file by overwriting the value with nothing.
>
> So yes, we could remove the line with the potential drawback of the webserver having a higher startup time and higher space consumption. Moving it to the environment file is not possible since (a) it is already there in form of the commented out line and (b) it would always affect both the consumer and the webserver.

pitkley:
But maybe we/@djmaze should add a comment to the YAML-file explaining that?

danielquinn:
I've added a comment as per @pitkley's recommendation, so I'm going to close this one now.

djmaze:
Thanks!

PR #96, opened 2016-03-21
Improves the docs: OCRing files in languages other than English + fixes typos

JaimeObregon:
Starts a "Troubleshooting" in the docs and fixes some typos.

### DIFF
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -8,7 +8,7 @@ should work) that has the following software installed on it:
 
 * `Python3`_ (with development libraries, pip and virtualenv)
 * `GNU Privacy Guard`_
-* `Tesseract`_, plus it's language files matching your document base.
+* `Tesseract`_, plus its language files matching your document base.
 * `Imagemagick`_
 * `unpaper`_
 
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -8,7 +8,7 @@ should work) that has the following software installed on it:
 
 * `Python3`_ (with development libraries, pip and virtualenv)
 * `GNU Privacy Guard`_
-* `Tesseract`_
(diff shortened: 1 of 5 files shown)


docs/requirements.rst
```diff
@@ -8,7 +8,7 @@ should work) that has the following software installed on it:
 
 * `Python3`_ (with development libraries, pip and virtualenv)
 * `GNU Privacy Guard`_
-* `Tesseract`_
+* `Tesseract`_, plus it's language files matching your document base.
```
danielquinn: Quick correction: this should be `its` rather than `it's`.

docs/troubleshooting.rst
```diff
@@ -0,0 +1,18 @@
...
+
+Consumer warns ``OCR for XX failed``
+------------------------------------
+
+If you find the OCR accuracy to be too low, and/or the document consumer warns that ``OCR for
+XX failed, but we're going to stick with what we've got since FORGIVING_OCR is enabled``, then you
+might need to install the `Tesseract language files
+<http://packages.ubuntu.com/search?keywords=tesseract-ocr>`_ marching your documents languages.
+
+As an example, if your documents are written in Spanish you may need to run::
```
danielquinn: We can't assume that our users are all using Debian or some derivitive thereof, so this should probably say something like:

> As an example, if your documents are written in Spanish and you're running a Debian-based system, you may need to run::

danielquinn:
I've been meaning to add a troubleshooting section, so thank you for posting this!  It's all good stuff, so if you can fix just those two bits, I'll merge it right away.

JaimeObregon:
Fixed! Note to self: do not submit pull request on Github when under sleep deprivation :-)

danielquinn:
Thanks for this!

Issue #102, opened 2016-03-30
Problem with Docker


buhino:
Hi guys, at first sorry for my english.

I have testing paperless in debian 8 jessie in my home and work's fine but in the work i have a problem with the permissions (Ubuntu 14.04).

docker-compose --version
docker-compose version 1.6.2, build 4d72027

I have done the same steps in Ubuntu:

1. git clone https://github.com/danielquinn/paperless.git
2. Set the permissions to 777 for avoid problems in the directory where I have installed Paperless and the consumer directory.

        aalcantara@ea:~$ ls -l
        drwxrwxrwx 2 aalcantara aalcantara     4096 mar 30 10:12 consumer
        drwxrwxrwx 3 aalcantara aalcantara     4096 mar 30 10:12 pruebas

  (Paperless is in the directory pruebas)

3. Modify the docker-compose.yml. in ./pruebas. It looks the following aspect:

        version: '2'
        
        services:
            webserver:
                image: pitkley/paperless
                ports:
                    # You can adapt the port you want Paperless to listen on by
                    # modifying the part before the `:`.
                    - "8000:8000"
                volumes:
                    - data:/usr/src/paperless/data
                    - media:/usr/src/paperless/media
                env_file: docker-compose.env
                # The reason the line is here is so that the webserver that doesn't do
                # any text recognition and doesn't have to install unnecessary
                # languages the user might have set in the env-file by overwriting the
                # value with nothing.
                environment:
                    - PAPERLESS_OCR_LANGUAGES=
                command: ["runserver", "0.0.0.0:8000"]
        
            consumer:
                image: pitkley/paperless
                volumes:
                    - data:/usr/src/paperless/data
                    - media:/usr/src/paperless/media
                    # You have to adapt the local path you want the consumption
                    # directory to mount to by modifying the part before the ':'.
                    - /home/users/aalcantara/consumer/:/consume
                    # Likewise, you can add a local path to mount a directory for
                    # exporting. This is not strictly needed for paperless to
                    # function, only if you're exporting your files: uncomment
                    # it and fill in a local path if you know you're going to
                    # want to export your documents.
                    # - /path/to/another/arbitrary/place:/export
                env_file: docker-compose.env
                command: ["document_consumer"]
        
        volumes:
            data:
            media:
 
Now the differences between my Debian 8 and Ubuntu 14.04:

Debian 8:

    artion@debian:/opt/paperless$ docker-compose up
    Creating network "paperless_default" with the default driver
[...]

PR #103, opened 2016-03-30
Don't fail on setting consumption-dir permissions

pitkley:
Changing the group or permissions on the consumption directory can fail, especially on network-mounted directories. If it fails, don't exit the consumer, but rather output a warning on what the user has to do.

This is related to (and should close) issue #102.


danielquinn:
I'm so glad you could figure out what was going wrong there.  I was fresh out of ideas after changing permissions to 777.

pitkley:
:+1:

PR #109, opened 2016-04-13
Catch inline attachments in email_fetcher

evils:
**Problem:**
* Paperless wouldn't import my emails send from my iPhone or Mac
* Behaviour was/is inconsistent, some emails get imported some not

A look into the source showed that the MailFetcher class would only check for "attachment" and therefore miss inline attachments as some email programs send them. Therefore added check to catch inline attachments via the "filename" disposition.

*Note: Never wrote python before and debugging was hard, make sure to check that single line of code that this PR changes.


danielquinn:
Since this didn't come with tests, I can't merge it right away.  I'll have to test it manually (and maybe write a proper unit test myself), but thank you for the PR!  I should get around to merging it later this week.

evils:
Hmm, I could have thought about that in first place! I'm about to write the Unit test, just a little challenged with python right now!

danielquinn:
Well if you do manage to include a test, I'd appreciate it.  There's already some tests in there based off of a test mail (also in there), so you could theoretically add a second test mail with an inline attachment and go from there.

Just remember to conform to pep8 (lines shouldn't exceed 80characters for example) and if TravisCI complains (as it has here) try to use that as a guide to see what went wrong.

Barring all that, do nothing and I'll get to this when I can ;-)  Either way, your contribution is appreciated as my experience parsing emails is limited.

danielquinn:
For example, your change might be better written as:

```python
if not dispositions[0].lower() == "attachment":
    if "filename" not in dispositions[1].lower():
        continue
```

This limits the need for parenthesis (the anti-Python!), makes things a little clearer, and keeps your line length under 79 characters (which is what TravisCI is complaining about above).

evils:
I've added UnitTests and took care of the line length issue.
Not sure if the line length issue is solved the way you want it to, but I don't like nested if's too much so I choose that option :)
The UnitTests aren't anything special, I had the same idea as what you mentioned: Copy your Unit Test!

My first OSS PR ... fingers crossed :+1:

danielquinn:
I'm so sorry that this took so long, but I've had a lot going on over here this past week.  I've merged your changes, with a few of my own.

First, congratulations on your first OSS PR! :-)  I'm glad that someone else is using the mail feature 'cause this area is *very* new water for me.  You also submitted tests, which was **super awesome** and makes merging it into master that much easier.  A few notes in case you were interested:

* The test you added was a copy/paste of my existing test for non-inline-mail.  That's cool, but there was a lot of duplication in there that didn't need to be.  So when I did the merge to master I also merged `test_mail_inline.py` with `test_mail.py`, just copying the relevant bits over.
* I hate backslashes in my code, but it's your fix and I hate it when people rewrite my Free software contributions so I'll leave it alone for now.  I can't promise it won't disappear in the future though ;-)

Thanks again for the PR!  It's always nice to see people using the stuff you build, but the collaborative process is even more fun.

PR #116, opened 2016-04-27
Update setup flow for bare metal with systemd

issmirnov:
The default files assume a specific path on the target system. Added a note letting people know they need to edit the files. Also fixed a bug with enabling the service - it caused odd errors on my system. Tested on Centos 7.

### DIFF
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -300,13 +300,19 @@ Standard (Bare Metal, Systemd)
 If you're running on a bare metal system that's using Systemd, you can use the
 service unit files in the ``scripts`` directory to set this up.  You'll need to
 create a user called ``paperless`` and setup Paperless to be in a place that
-this new user can read and write to.  Then, you can just tell Systemd to enable
-the two ``.service`` files::
+this new user can read and write to. Be sure to edit the service scripts to point
+to the proper location of your paperless install. Note that if your system aliased
+``/usr/bin/python`` to python3 you should modify the ``ExecStart`` line to look like
+this: ``ExecStart=/usr/bin/python3 /path/to/paperless/src/manage.py document_consumer``.
+If you don't want to make a new user, you can change the ``Group`` and ``User`` variables
+accordingly.
 
-    # systemctl enable /path/to/paperless/scripts/paperless-consumer.service
-    # systemctl enable /path/to/paperless/scripts/paperless-webserver.service
-    # systemctl start /path/to/paperless/scripts/paperless-consumer.service
-    # systemctl start /path/to/paperless/scripts/paperless-webserver.service
+Then, you can just tell Systemd to enable the two ``.service`` files::
+
+    # sudo systemctl enable /path/to/paperless/scripts/paperless-consumer.service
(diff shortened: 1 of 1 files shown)


docs/setup.rst
```diff
@@ -300,13 +300,19 @@ Standard (Bare Metal, Systemd)
...
-the two ``.service`` files::
-
-    # systemctl enable /path/to/paperless/scripts/paperless-consumer.service
-    # systemctl enable /path/to/paperless/scripts/paperless-webserver.service
-    # systemctl start /path/to/paperless/scripts/paperless-consumer.service
-    # systemctl start /path/to/paperless/scripts/paperless-webserver.service
+this new user can read and write to. Be sure to edit the service scripts to point
+to the proper location of your paperless install. Note that if your system aliased
+``/usr/bin/python`` to python3 you should modify the ``ExecStart`` line to look like
+this: ``ExecStart=/usr/bin/python3 /path/to/paperless/src/manage.py document_consumer``.
```
pitkley: If `python` is aliased to `python3`, then why change the `ExecStart`? Do you mean if `python` is aliased to `python2`?

docs/setup.rst
```diff
@@ -300,13 +300,19 @@ Standard (Bare Metal, Systemd)
...
+If you don't want to make a new user, you can change the ``Group`` and ``User`` variables
+accordingly.
+
+Then, you can just tell Systemd to enable the two ``.service`` files::
+
+    # sudo systemctl enable /path/to/paperless/scripts/paperless-consumer.service
+    # sudo systemctl enable /path/to/paperless/scripts/paperless-webserver.service
+    # sudo systemctl start paperless-consumer
+    # sudo systemctl start paperless-webserver
```
pitkley: The `#`-sign generally indicates that you have to execute the line as `root`. Rather than specifying `sudo` here, changing the line above to something like:

> Then, you can just tell Systemd as `root` (or using `sudo`) to enable the two `.service` files::

might be the better option.

docs/setup.rst
danielquinn: I agree with @pitkley on this one.  The way it's worded here makes it more confusing.  I do however agree that we should have a note here explaining that the `.service` files will likely need to be edited to fit your environment.  However all that should be required is something like:

> Be sure to edit the `.service` scripts to point to the proper location of your paperless install, referencing the appropriate Python binary.  For example:
> `ExecStart=/path/to/python3 /path/to/paperless/src/manage.py document_consumer`

issmirnov:
Thanks for the great feedback. I've added a second commit to address your comments. How does it look now?

issmirnov:
Oops, I accidentally closed this instead of simply commenting. Reopening.

danielquinn:
Looks great!  Thanks for your contribution :-)

issmirnov:
Of course, thank _you_ for a wonderful project! This PR was generated as a side effect of me setting this up on our home server. Going to look at our scanner and see if it can do email well, and then looking forward to shredding the entire file cabinet.

PR #124, opened 2016-05-14
Density Setting

bmartin5692:
This provides for a new optional configuration setting for density. If it isn't specified in the config, consumer defaults to 300 like normal.
A blurb has been added to the sample config regarding the setting.


paperless.conf.example
```diff
@@ -61,6 +61,17 @@ PAPERLESS_SHARED_SECRET=""
...
+# By default the conversion density setting for documents is 300DPI, in some
+# cases it has proven useful to configure a lesser value.
+# This setting has a high impact on the physical size of tmp page files,
+# the speed of document conversion, and can affect the accuracy of OCR
+# results. Individual results can vary and this setting should be tested 
+# thoroughly against the documents you are importing to see if it has any 
+# impacts either negative or positive. Testing on limited document sets has
+# shown a setting of 200 can cut the size of tmp files by 1/3, and speed up
+# conversion by up to 4x with little impact to OCR accuracy.
+#PAPERLESS_CONVERT_DENSITY=200
```
danielquinn: Everything looks good, but if the default is 300, then the commented-out value for `PAPERLESS_CONVERT_DENSITY=` should also be 300 so as not to confuse people.  So if you change this to 200, I can merge it.

bmartin5692: Sounds good. Changes have been made.

danielquinn:
Thanks for adding this, and for making that quick fix I asked for!

PR #125, opened 2016-05-30
django-filter 0.12.0 no longer available

Djailla:
While installing paperless using pip, django-filter failed to install 0.12.0 version and now 0.13.0 is available

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Django==1.9.4
 Pillow==3.1.1
 django-crispy-forms==1.6.0
 django-extensions==1.6.1
-django-filter==0.12.0
+django-filter==0.13.0
 djangorestframework==3.3.2
 filemagic==1.6
 langdetect==1.0.5


pitkley:
```console
root@5a1822f47d0c:/# pip install django-filter==0.12.0
Collecting django-filter==0.12.0
  Downloading django_filter-0.12.0-py2.py3-none-any.whl
Installing collected packages: django-filter
Successfully installed django-filter-0.12.0
```

Also [still available on PyPi](https://pypi.python.org/pypi/django-filter/0.12.0). So what you experience seems to be an issue on your end, or was just something intermittent.

Checking `django-filter`s [changelog](https://github.com/carltongibson/django-filter/blob/develop/CHANGES.rst#version-0130-2016-03-11), upgrading shouldn't be an issue though.

danielquinn:
@Djailla, were you able to install django-filter 0.13.0?  As @pitkley pointed out, Paperless shouldn't have any trouble running with an updated version, but I don't like bumping up the version numbers of requirements unless they're actually needed.

Djailla:
@danielquinn yes I was able to install django-filter 0.13.0. Agreed on the fact to not update requirements if not needed.

Just for the information I'm trying to port paperless in NAS OS Seagate App (based on this SDK : http://www.seagate.com/nasos/SDK/0.7/). The container used is Debian Wheezy, I'll check tomorrow if it's systematic.

danielquinn:
I hadn't realised that I'd left this PR open.  What say you to changing this line to:

    django-filter>=0.12.0

Djailla:
That would be good ;)

danielquinn:
Alright, I modified `requirements.txt` to allow most packages to upgrade (everything but Django, since they're unlikely to break things while Django likely would).  As my change voids this PR, I'm going to mark this as closed now.

Thanks for pointing this out, and I'm sorry I didn't respond to it sooner.

Issue #128, opened 2016-06-24
Paperless not waiting for scan to be completely transmitted


phryneas:
When I'm doing a big scan (in this specific case, 21 pages and 7MB), paperless starts processing it before it is completely copied to the consumption folder.

I'm helping myself with a pre-consumption script like this right now:

    /bin/bash
    while [ $(( $(date +%s) - $(stat -c %Y $1) )) -lt 10 ]; do sleep 1; done;

is there an option that already does this or could something like this be implemented?

danielquinn:
Actually, this is a bug.  [There's code in the consumer to account for this](https://github.com/danielquinn/paperless/blob/master/src/documents/consumer.py#L103) but it's clearly not working.  Currently it's doing a simple test for file size and then re-testing it on the next loop.  If the sizes are the same, then it's assumed that it can be consumed.

...at least, that's what it's supposed to be doing.  [Have a look yourself](https://github.com/danielquinn/paperless/blob/master/src/documents/consumer.py#L339) if you think I've missed something because this isn't the first time this has happened.  It's difficult to write a test for, so I've not been able to track down the problem yet.

pitkley:
@danielquinn Isn't the check [you mentioned](https://github.com/danielquinn/paperless/blob/master/src/documents/consumer.py#L103) the wrong way around? If I understand it correctly, the loop skips (`continue`) over a document if `_is_ready` is `True`, instead of `False`.

danielquinn:
Heh, I think you're right.  @phryneas can you change this to `if not self._is_ready(doc):` and see if it works for you?  I've spent all this time thinking that there was a problem with the method that I never took a second look at the calling of it.

phryneas:
Funny... I read your code a few hours ago, didn't really understand it (my python sucks) and then went offline. An hour or so later still thinking about it, it occured to me that the check might be backwards xD

I'll give it a try :+1:

phryneas:
I did a few tests, the negated check seems to work like a charm for me. Thanks :)

danielquinn:
Hooray!  If you submit the fix as a pull request, you'll get credit for the fix.  Otherwise, I'll just do it.

phryneas:
not gonna deny a free contribution xD

there you are

PR #129, opened 2016-06-26
wait until file is completely transmitted

phryneas:
negation was missing for feature to be active, see #128


danielquinn:
Thanks!  This one was bugging me for a long time.

PR #137, opened 2016-08-16
Upgrade Django to 1.10

JamieMagee:
This includes a security patch for CVE-2016-6186


danielquinn:
Looks good, thanks for the PR :-)

PR #148, opened 2016-08-23
Allow database location to be variable

jbogatay:


src/paperless/settings.py
```diff
@@ -90,9 +95,10 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.sqlite3",
-        "NAME": os.path.join(BASE_DIR, "..", "data", "db.sqlite3"),
+        "NAME": os.path.join(os.getenv("PAPERLESS_DBLOCATION",os.path.join(BASE_DIR, "..", "data")), "db.sqlite3")
```
danielquinn: Nitpick: you've got a couple of pep8 violations here:

* The line is too long (should be less than 80 characters)
* There's no space between `,` and `os.`.  Maybe something like this instead?

```python
os.path.join(
    os.getenv(
        "PAPERLESS_DBLOCATION",
        os.path.join(BASE_DIR, "..", "data")
    ), 
    "db.sqlite3"
)
```

Additionally, if `PAPERLESS_DBLOCATION` is set to the value of the database *file* wouldn't you end up with something like `/path/to/db.sqlite3/db.sqlite3`?

Maybe this would be better?

```python
os.getenv(
    "PAPERLESS_DBLOCATION",
    os.path.join(BASE_DIR, "..", "data", "db.sqlite3")
)
```

Unless of course you meant to have `PAPERLESS_DBLOCATION` set to just the directory?  In which perhaps the variable name should be `PAPERLESS_DB_DIRECTORY` just so it's obvious?

danielquinn: Strike that, my suggestion would be `PAPERLESS_DBDIR` to remain consistent with the other _DB* names.  I just think that "DIR" should be there just to make it obvious.

jbogatay:
I'll fix it up.   Dunno about python conventions, this is my first anything with python.   I consider it self-therapy for my whitespace sensitive language aversion.

Location implies... location to me.  Not file.   But dir is cool too.

jbogatay:
Also, I plan on doing MEDIA_ROOT next, and then at some point a refactor of the Dockerfile.  900+GB overhead is way too much.

danielquinn:
Excellent, and thanks for making the requested changes.  My Docker-foo is still quite limited, so have at it.  I've been working my other projects using `python:3.5-alpine`, so if you can get this to work with that, it'd be handy.

If you tinker with the Docker stuff though, you may want to talk to @pitkley as he's been handling the Docker stuff so far.  I'm sure you two can work together to make it better :-)

As for enabling reassignment of the `MEDIA_ROOT` values, go nuts.  So long as the defaults still work and everyone isn't affected by an upgrade, I'm cool with it.

jbogatay:
Yes, alpine was what I had in mind.   And perhaps fewer layers.   I retooled some other containers that I publish and the size reduction was 90%.   Override default locations in the container so they are prettier.  (/app/data -vs- /usr/src/....).    Container size is just a peeve of mine, lol.   1.2MB of code, 900MB of cruft.

Going to add a toggle for inlining the documents (Content-Disposition: inline).   That file prompt drives me batty, especially since the major browsers can show most content just fine.

Also, might add a 'date-added' type field.   While I am tweaking my tags and such, I'd like to see the last X added documents for review.  My doc dates/scan dates are all over the map, and frequently newly added documents are buried in the list somewhere.

danielquinn:
Those all sound like good changes, just please keep them as separate PRs so we can manage the merging easier.

PR #150, opened 2016-08-23
Override media_root option

jbogatay:
Sorry about the re-send.   Changed my workflow a little bit.

### DIFF
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -151,7 +151,10 @@ USE_TZ = True
 # https://docs.djangoproject.com/en/1.9/howto/static-files/
 
 STATIC_ROOT = os.path.join(BASE_DIR, "..", "static")
-MEDIA_ROOT = os.path.join(BASE_DIR, "..", "media")
+MEDIA_ROOT = os.getenv(
+                "PAPERLESS_MEDIADIR",
+                os.path.join(BASE_DIR, "..", "media")
+             )
 
 STATIC_URL = '/static/'
 MEDIA_URL = "/media/"
--- a/paperless.conf.example
+++ b/paperless.conf.example
@@ -85,3 +85,5 @@ PAPERLESS_SHARED_SECRET=""
 # the default location
 #PAPERLESS_DBDIR=/path/to/database/file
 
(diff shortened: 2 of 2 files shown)


src/paperless/settings.py
```diff
@@ -151,7 +151,10 @@
 # https://docs.djangoproject.com/en/1.9/howto/static-files/
 
 STATIC_ROOT = os.path.join(BASE_DIR, "..", "static")
-MEDIA_ROOT = os.path.join(BASE_DIR, "..", "media")
+MEDIA_ROOT = os.getenv(
```
danielquinn: Another nitpick: pep8 rules say that all indents must be with 4  spaces.  I suggest this instead:

```python
MEDIA_ROOT = os.getenv(
    "PAPERLESS_MEDIADIR", os.path.join(BASE_DIR, "..", "media"))
```

or

```python
MEDIA_ROOT = os.getenv(
    "PAPERLESS_MEDIADIR",
    os.path.join(BASE_DIR, "..", "media")
)
```

jbogatay:
What do you use to run the pep8 checks?   I'm having a hard time finding something that jives.

- After editing tox so pep8 command line runs against settings, it only complains about excessive blank lines and long lines elsewhere.
- online pep8 is same as above.
- sublime plugin wants to change it to something totally different than either of your suggestions.

danielquinn:
Honestly, I was just eyeballing it, which I think was the problem.

As since your own tests were saying everything was fine, I checked out the code locally and looked at it, running it with `pep8 src/paperless/settings.py` just to be sure.  You're right, it passes just fine.  The indentation of the two middle lines is a multiple of 4, and for some reason, pep8 isn't concerned about the hanging `)` which is indented with 13.

So my bad, this is *not* a pep8 violation, it's just *weird* :-)  Typically you don't see a closing bracket align to the opening line like that, but rather things are indented with one indentation greater than the line the statement started on (as in my samples above).  But you're right, it's legit code, so I'll stop messing with you and merge it :-)  I may just have to change it to conform to my preferences later or risk setting my OCD off.

jbogatay:
I'm cool with fixing weird.   Not sure what's normal pythoneese yet.

PR #152, opened 2016-08-26
Update setup.rst

RubenWaterman:
Added a line on how to use the ubuntu upstart example in Vagrant. It took me hours to figure out why it wouldn't start on boot but would when you would type "start paperless-server" manually. The start on vagrant-mounted solved the problem, so I thought it would be good to share this with the paperless community.


danielquinn:
This is exactly the kind of PR I like to receive!  Thanks for the update!

PR #159, opened 2016-10-05
Updating Django admin event log on document_consumption_finished

synchrone:


danielquinn:
I had to make some changes on this one, specifically, I moved the logging logic into the consumer and out of the management command.  The code you had in there set a receiver on a class method, sending `self` as the sender and I'm not convinced it would work properly.  Regardless, the change I did here made the code a lot smaller.  I hope it still solves your problem, but if it doesn't please say so here and I'll see what I can do.

Issue #88, opened 2016-03-17
Better text layout on OCR scrapped information 


ralesi:
I notice when pulling text from PDFs with pdftk,  we are able to obtain whitespace formatted text which gives a rough visual representation of the PDF.   Is it possible to do this with the current tools used to OCR the PDF files created with paperless? 

Thanks for the great work!

danielquinn:
This is a common question, and unfortunately, I think (correct me if I'm wrong) the ability to use pdftk to pull out text along with formatting is *limited to software-generated PDFs* like what you get when you save an office document as a PDF.

In Paperless' case, the PDFs coming in are just rasterised images sent from a document scanner.  These files don't have a text layer embedded in them, so we have to use an OCR program (Tesseract) to "read" the document and parse out the text.  However, what comes out of this process isn't a neatly formatted text, but rather a long string of text that is the OCR reader's best guess.  I'm afraid I know of no way to pull out all of the text in a neatly formatted fashion.  It may exist of course, but I don't know what it is :-(

danielquinn:
This question is related to #19, so if you're curious about what other people might know on this front, that might be a good place to start.

ralesi:
Thanks,  looking at stack overflow,  looks like tesseract has a new option : `preserve_interword_spaces`.  Have we given this a try?

danielquinn:
Hmm.  Nope.  I'll give it a shot and see what happens :-)

danielquinn:
Ok, so I've done some poking around and there's good news and bad:

#### The Bad News

Adding that flag (`tesseract -l eng -c preserve_interword_spaces=1 src/test.pnm -`) doesn't seem to have much of a positive effect when compared to the default (`tesseract -l eng src/test.pnm -`).  It doesn't seem to hurt though, so I can see about enabling it.

#### The Good News

Tesseract's default output includes (at the bare minimum) newline characters (`\n`) where appropriate, greatly contributing to readability.  *By default, I'm currently stripping them*, along with any extra spaces to make searching for text strings easier.  If preserving this white space is important, I can either (a) store the stuff twice (once formatted, once stripped) or (b) figure out how to make searches work against these spaced-out text fields.

So, in summary: I now have an idea how to fix this and it's on my list :-)

synchrone:
@danielquinn I have modified the regex you're using to strip the excessive whitespace, but retain the newline characters for readability, all in one field, like so:
```python
[...]

PR #160, opened 2016-10-11
Collapsing excess whitespace after OCR, but keeping newlines

synchrone:
Issue #88 calls for treating OCRed text like actual data, not just search index. 
This PR makes it retain newlines while still collapsing all the multiple whitespace that tesseract returns into one space character.


danielquinn:
Sorry about the delay, but I had to fix the pep8 violations.  Thanks for the PR!

PR #163, opened 2016-10-23
Allow to create case sensitive matches

CkuT:


danielquinn:
There was a pretty ugly ValueError in this PR where you passed `search_kwargs` to `re.search()` using a compiled regex.  This is what was upsetting Travis (along with a few pep8 violations).  I've fixed both and merged it, so do a `git pull` on your master branch and you should be up to date now.

In the future, please remember to run the tests so you can catch this stuff early.  I've now modified the `requirements.txt` file to include the test modules so all you should need to do is:

```bash
pip install -r requirements.txt
tox
```

Tox may complain about missing versions of Python, don't worry about that, you only need one of them to run the tests.  Alternatively, you can run `pytest` instead of `tox` which will run the Python tests and `pep8` to run the pep8 test.  Just ignore anything complaining about `settings.py` or anything in `.tox` or `migrations`.

PR #174, opened 2017-01-04
Randomize secret key

### LINKED ISSUE #171: Dynamic Secret Key Base
Hiho, you're setting the secret key statically, it would be nice to load them from environment or config.
https://github.com/danielquinn/paperless/blob/master/src/paperless/settings.py#L25

dopykuh:
Provide ability to load secret key from environment otherwise use a random uuid. It shouldn't affect older versions. Closes #171

### DIFF
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -11,6 +11,7 @@ https://docs.djangoproject.com/en/1.9/ref/settings/
 """
 
 import os
+import uuid
 
 from dotenv import load_dotenv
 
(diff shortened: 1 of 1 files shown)


danielquinn:
There's a couple things that have to be fixed here before your change can be merged:

1. The change you've got here essentially randomises the secret key on every server start.  This would mean that any time Paperless is upgraded & restarted, or the server restarts, or something crashes, all of the data verified against the secret key (like logins) will be invalidated.  A better choice would be something like `os.getenv("PAPERLESS_SECRET_KEY", "e11fl1oa-*ytql8p)(06fbj4ukrlo+n7k&q5+$1md7i+mge=ee")`
2. When you introduce a new configurable value, it's polite to include it in `/paperless.conf.example` so that people will see it when doing their initial setup, along with a description as to what this value is for and why someone may want to set it.  I'm referring to [something like this](https://github.com/danielquinn/paperless/commit/7b586e68579c1ef84efe1c3076c7edecfa89142b#diff-9921e14d1c87f420c220c2933c630128).

If you can do those things, I'll merge it tonight :-)

danielquinn:
I've implemented this myself using the above method, so I'm gonna close this.

Issue #172, opened 2017-01-02
Possible to specify ALLOWED_HOSTS in paperless.conf?


ekw:
My paperless installation doesn't run unless I add my hostname to ALLOWED_HOSTS in 'src/paperless/settings.py'.  Rather than modifying that source file, can I specify it in '/etc/paperless.conf' instead?  Since settings.py has load_dotenv("/etc/paperless.conf") statement in it, I tried putting these (and probably a few more variations) in /etc/paperless.conf with no success:
```
ALLOWED_HOSTS=['my_hostname']
ALLOWED_HOSTS='my_hostname'
ALLOWED_HOSTS=my_hostname
PAPERLESS_ALLOWED_HOSTS=['my_hostname']
PAPERLESS_ALLOWED_HOSTS='my_hostname'
PAPERLESS_ALLOWED_HOSTS=my_hostname
```
Do I just not have the correct syntax or is this not possible?  Thanks.

danielquinn:
ALLOWED_HOSTS is only necessary if you're putting Paperless in a public environment and intend to turn `DEBUG` off.  You can set it, but since I never intended Paperless to go on a public machine, I haven't prepared the code for that.

The problem you're having is that if you set `ALLOWED_HOSTS=os.getenv("PAPERLESS_ALLOWED_HOSTS")` what you're telling Python is to set `ALLOWED_HOSTS` to the *string* value of whatever is in `${PAPERLESS_ALLOWED_HOSTS}`.  Since Django expects this variable to be a list, that will never work.

So in other words, you're actually doing this:

```python
ALLOWED_HOSTS="['my_hostname']"
```

The work-around for this is kind of a pain in the ass because of how Python works, but it would look something like this:

```bash
# /etc/paperless.conf
PAPERLESS_ALLOWED_HOSTS=example.com
```

```python
ALLOWED_HOSTS = []

allowed_hosts = os.getenv("PAPERLESS_ALLOWED_HOSTS")
if allowed_hosts:
    ALLOWED_HOSTS = allowed_hosts.split(",")
```

This would allow you to not define PAPERLESS_ALLOWED_HOSTS, or do stuff like this:

```bash
# /etc/paperless.conf
PAPERLESS_ALLOWED_HOSTS=example.com
```

or

```bash
# /etc/paperless.conf
PAPERLESS_ALLOWED_HOSTS=example.com,example.net
```

Either that, or I could just hard-code this:

```python
ALLOWED_HOSTS=["*"]
```

[...]

Issue #176, opened 2017-01-06
Make OCR_LANGUAGE configurable


danielquinn:
I can't believe I didn't make this configurable earlier.  Oops.

PR #177, opened 2017-01-07
Show document thumbnails; Use django-suit theme

ekw:
Continuing our discussion from https://github.com/danielquinn/paperless/issues/172, here's the pull request.  I tried the git reset thing, but I managed to somehow make a bigger mess.  There weren't many changes, so I nuked my fork and started from scratch. =)


ekw:
I don't know how my 2nd commit "Make timezone configurable" made it into this pull request.  I created the pull request only for the 1st commit, bee0867.  But when I later committed e6526d3, it became a part of this pull request automatically.  But it is now there for your further consideration!

denysvitali:
OT: I love how git can be both powerful and destructive at the same time

ekw:
Would you be interested in a pull request for making list_per_page configurable?  I want to customize that without modifying admin.py or settings.py.  I can tackle #176 at the same time.

danielquinn:
Pull requests are tied to a branch, not a commit, so if you do a pull request on a branch, and then commit & push updates to that branch, they become part of that pull request. 

That's why typically you create a feature branch and do a pull request from that branch.  That'll let you create a separate branch for different features without worrying about mucking up the first PR.

As for making list_per_page configurable, sure I don't have any objections.  You can include it in this PR or make a separate one. 

My computer is off for the night and I'm on my phone right now, so I'll have a look at whatever you posted tomorrow.

paperless.conf.example
```diff
@@ -102,3 +102,13 @@ PAPERLESS_SHARED_SECRET=""
 # as is "example.com,www.example.com", but NOT " example.com" or "example.com,"
 #PAPERLESS_ALLOWED_HOSTS="example.com,www.example.com"
 
+# Override the default UTC time zone here
+#PAPERLESS_TIME_ZONE=America/Los_Angeles
```
danielquinn: I love the idea of making the timezone configurable, but the default is UTC, so that should be what we have here, commented out so the user knows their starting position.

danielquinn:
It all looks good!  There's a couple places where I'll want to fiddle with the indentation because I'm picky like that, but I won't force you to do it :-)  I just need that one comment about the timezone fixed and I'll merge it.

danielquinn:
All merged!  Thanks so much for your addition!  Django suit really does make things *so* much nicer.

Ulli2k:
I was hoping to get thumbnails for a long time ago. Thanks a lot for implementing this!
Just a question because I think with that kind of view we are loosing the overview of available documents.
Would it be possible to show the thumbnails of each document in case of a "mouse-over" or a click on a document link/icon?

danielquinn:
This isn't all that difficult, but at the moment, [we're discussing whether we can continue using django-suit](https://github.com/danielquinn/paperless/issues/180#issuecomment-271249081) (#180) so give me some time to sort that out, and perhaps I can find a way to make the ui work for everyone.

Issue #178, opened 2017-01-08
Allow HTTP Basic auth for /fetch endpoints


thomasbrueggemann:
Probably by pulling the /fetch functionality under the DjangoRestFrameworks scope.
Currently the FetchView can only be reached via SessionAuthentication login.

_I am new to Django but would love to contribute to this instead of just nagging about it 🙂_

danielquinn:
Well diving into DjangoRestFramework may not be the easiest way to learn Django, but I'm more than happy to let you try :-)  If you're reasonably adept at Python, it shouldn't be too much to handle.

I'll assign this one to you, and feel free to post questions here if you have any.

PR #179, opened 2017-01-08
The /fetch endpoint can be authenticated via session or via HTTP basic

thomasbrueggemann:
I wasn't able to "pull the /fetch endpoint under the Django RestFramework" as proposed in the issue #178 but found a great way to check via a Mixin if the session or HTTP Basic auth was used to call the /fetch and /push endpoint.

The Django RestFramework magic is definitly to much to read into in reasonable time.
What do you think about it, @danielquinn ?

### DIFF
--- a/src/documents/mixins.py
+++ b/src/documents/mixins.py
@@ -48,8 +48,8 @@ class SessionOrBasicAuthMixin(AccessMixin):
                             login(request, user)
                             request.user = user
                             return super(
-								SessionOrBasicAuthMixin, self
-							).dispatch(
+                                SessionOrBasicAuthMixin, self
+                            ).dispatch(
                                 request, *args, **kwargs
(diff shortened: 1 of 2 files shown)


danielquinn:
Porting the FetchView to DRF would probably be more elgant, but I'm not in a hurry to make things difficult just for two views.  I'm fine with this setup, so long as you can get it to conform to pep8 (that's they the check failed: you've got a 3-space indent in your class there).

So, if you can fix that, I'll merge the change.

thomasbrueggemann:
Working on the pep8 fix right now.

thomasbrueggemann:
Done.

danielquinn:
Nice!  Merged!

thomasbrueggemann:
Awesome, thanks!

PR #181, opened 2017-01-10
Remove django-suit

ekw:


danielquinn:
Thanks for doing this.  I've been swamped these past few days.

PR #184, opened 2017-01-23
Include responsive/mobile UI

elohmeier:
Hi @danielquinn,
there is a nice small package around, called [django-flat-responsive](https://github.com/elky/django-flat-responsive). Since you decided not to use django-suit for now, this improves the stock django admin experience on mobile browsers. I use it for browsing my documents on the go.

Regards
Enno

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,6 +12,7 @@ python-dotenv>=0.3.0
 python-gnupg>=0.3.8
 pytz>=2015.7
 gunicorn==19.6.0
+django-flat-responsive==1.2.0
 
 # For the tests
 pytest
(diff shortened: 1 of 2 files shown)


danielquinn:
This sounds reasonable enough, but I'll want to tinker with it myself for a bit and double-check the license before I just merge it.

danielquinn:
That package is remarkably clean and simple.  I like it.  Merging now!  Sorry about the delay and thanks for the PR :-)

elohmeier:
FYI, starting with Django 2.0 this package is not needed anymore as the stock Django admin is responsive already.

danielquinn:
Yeah, I've been meaning to upgrade to 2.x, but there were some minor headaches with the way URL lookups changed, so I've been putting it off.  When I do make the switch though, I'll definitely remove this depedency.  Thanks for the heads-up :-)

Issue #120, opened 2016-05-03
Re-Tagger for Correspondents!? 


cwasher:
Installed paperless as a docker instance and loving it so far. I have scanned 100s of documents and have a box of "to-be shred" filling up nicely! 

I like the tag feature and the ability to reapply tags to existing documents, does a similar script exist for Correspondents? Or did I miss something? 

Thanks!

danielquinn:
Heh.  Nope.  That doesn't exist yet because the re-tagger was written after tags were invented, but *before* auto-correspondents were done.  I'll tag this one as an `enhancement`.

stgarf:
@cwasher It looks like this issue may have been resolved by #185. Would you mind taking a look and confirming and then closing out the issue? Thanks!

rjendoubi:
A year on, I too would suggest this has been covered and can be closed :smiley:

PR #185, opened 2017-01-31
Add manager command to re-tag documents without correspondent

rhaamo:
hello,

Like in #120 I've needed this possibility so I've created a "document_correspondents" manager command like the one for tags.
I does ignore if a correspondent is already associated to a document.
I've based the re-tagger command for correspondent on the code from the signal/handler. 

I tried it with a (small) set of files with either regexp or litteral and it does seems to work as intended.

### DIFF
--- a/src/documents/management/commands/document_correspondents.py
+++ b/src/documents/management/commands/document_correspondents.py
@@ -6,12 +6,11 @@ from ...mixins import Renderable
 
 
 class Command(Renderable, BaseCommand):
-
     help = """
         Using the current set of correspondent rules, apply said rules to all
         documents in the database, effectively allowing you to back-tag all
-        previously indexed documents with correspondent created (or modified) after
-        their initial import.
+        previously indexed documents with correspondent created (or modified)
+        after their initial import.
     """.replace("    ", "")
 
     def __init__(self, *args, **kwargs):
@@ -27,7 +26,8 @@ class Command(Renderable, BaseCommand):
             if document.correspondent:
                 continue
 
-            potential_correspondents = list(Correspondent.match_all(document.content))
+            potential_correspondents = list(
+                Correspondent.match_all(document.content))
             if not potential_correspondents:
                 continue
 
@@ -35,9 +35,11 @@ class Command(Renderable, BaseCommand):
 
             selected = potential_correspondents[0]
             if potential_count > 1:
-                message = "Detected {} potential correspondents for {}, so we've opted for {}"
+                message = "Detected {} potential correspondents for {}, " \
+                          "so we've opted for {}"
                 print(message.format(potential_count, document, selected))
 
-            print('Tagging {} with correspondent "{}"'.format(document, selected))
+            print('Tagging {} with correspondent "{}"'.format(document,
+                                                              selected))
             document.correspondent = selected
             document.save(update_fields=("correspondent",))
--- a/src/documents/management/commands/document_correspondents.py
+++ b/src/documents/management/commands/document_correspondents.py
@@ -1,0 +1,43 @@
+from django.core.management.base import BaseCommand
+
+from documents.models import Document, Correspondent
+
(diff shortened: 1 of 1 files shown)


danielquinn:
I'll have to spend a little time playing with this before I merge it, but at first glance it looks good.  I've got a conference to do over the weekend, but I'll try to take a look before Monday.

src/documents/management/commands/document_correspondents.py
```diff
@@ -0,0 +1,45 @@
...
+                continue
+
+            potential_correspondents = list(
+                Correspondent.match_all(document.content))
+            if not potential_correspondents:
+                continue
+
+            potential_count = len(potential_correspondents)
+
+            selected = potential_correspondents[0]
```
danielquinn: I'm not thrilled about this decision here.  Imagine a situation where a user had 100 files that they haven't yet given a correspondent to.  They then create new correspondent matching rules for both their bank and for their ISP and run this script only to realise that they have their bank name on all of their ISP's invoices and now everything has be arbitrarily labelled with the wrong correspondent and there's no easy way to fix it.

A suggestion:

Why not draw a simple text menu + select option for cases where multiple correspondents match?  I'm thinking of something like this:

```
The file http://localhost:8000/fetch/doc/282 matches multiple potential correspondents.  Which of the following would you like to assign to it?

0. None of these, leave it empty
1. Telus Mobility
2. Scotiabank

Your selection[0]: 
```

Additionally, I'd like to have some sort of standardised log output for what goes down should the user screw up so they can then use that log output as a way to batch-fix whatever they've done.  Something like this:

```
Document 282: "ISP Bill - utilities.pdf" was given the correspondent id 7: "Telus Mobility"
```

src/documents/management/commands/document_correspondents.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, *args, **kwargs):
+        self.verbosity = 0
+        BaseCommand.__init__(self, *args, **kwargs)
+
+    def handle(self, *args, **options):
+
+        self.verbosity = options["verbosity"]
+
+        for document in Document.objects.all():
+            # No matching correspondents, so no need to continue
```
danielquinn: This comment should probably be something like:

```python
# The document already has a correspondent. Skipping.
```

src/documents/management/commands/document_correspondents.py
CkuT: Won't it be a little annoying for the user to specify the right correspondent for every file ?

Maybe add an option to leave it empty by default if there are multiple matched correspondent ?

src/documents/management/commands/document_correspondents.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self.verbosity = 0
+        BaseCommand.__init__(self, *args, **kwargs)
+
+    def handle(self, *args, **options):
+
+        self.verbosity = options["verbosity"]
+
+        for document in Document.objects.all():
```
CkuT: If you are going to skip `Documents` with a correspondent, maybe a `Document.objects.filter(correspondent=None)` would be better ?

CkuT:
Any updates on this ? If some help is needed, I'm here :)

danielquinn:
Sorry about the delay on this guys.  It's the sort of thing I want to be sure works Just Right™ and I'm in the process of getting married, so Paperless has taking a back-burner position of late.  I'll get to this as soon as I can.

maur:
@danielquinn any news, if you're not too busy?

danielquinn:
Hi guys, I'm so sorry this took as long as it did (looks like almost exactly a **year**), but thanks to a last-ditch effort from @maur to poke me, I had some time tonight and I merged this manually after making a few changes and adding a `--use-first` flag.  Go ahead and give it a try and let me know what you think.

PR #186, opened 2017-01-31
Be able to override static root using config file

rhaamo:
Like PAPERLESS_MEDIADIR this add option to also override STATIC_ROOT.

My setup use an external directory for STATIC files created using _manager.py collectstatic_ and this would make maintaining paperless sources easier by having settings.py untouched.


danielquinn:
This is very clean code, thank you :-)

PR #194, opened 2017-02-28
Better alt-text for thumbnails.

philippeowagner:


danielquinn:
Nice catch, thanks for the PR!

PR #197, opened 2017-03-11
Pluggable consumers

danielquinn:
> **I'm setting this up as a pull request to see if there are any comments before I just go ahead and merge it.  I'll probably leave it as a PR for about a week, so if you have ideas/criticism, I'm all ears.**

I've broken out the OCR-specific code from the consumers and dumped it
all into its own app, `paperless_tesseract`.  This new app should serve
as a sample of how to create one's own consumer for different file
types.

Documentation for how to do this isn't ready yet, but for the impatient:

* Create a new app
    * containing a `parsers.py` for your parser modelled after
      `paperless_tesseract.parsers.RasterisedDocumentParser`
    * containing a `signals.py` with a handler moddelled after
      `paperless_tesseract.signals.ConsumerDeclaration`
    * connect the signal handler to
      `documents.signals.document_consumer_declaration` in
      `your_app.apps`
* Install the app into Paperless by declaring
  `PAPERLESS_INSTALLED_APPS=your_app`.  Additional apps should be
  separated with commas.
* Restart the consumer


philippeowagner:
Had no chance yet to dive into the details/code but the description alone earns a thumbs up!

PR #212, opened 2017-04-08
Docker: prepare export directory

Strubbl:
This commits prepares an /export directory to be used for the paperless exporter

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,6 @@ RUN groupadd -g 1000 paperless \
 # Set export directory
 ENV PAPERLESS_EXPORT_DIR /export
 RUN mkdir -p $PAPERLESS_EXPORT_DIR
-RUN chown -Rh paperless:paperless $PAPERLESS_EXPORT_DIR
 
 # Setup entrypoint
 COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,12 +35,17 @@ RUN groupadd -g 1000 paperless \
     && useradd -u 1000 -g 1000 -d /usr/src/paperless paperless \
     && chown -Rh paperless:paperless /usr/src/paperless
 
+# Set export directory
+ENV PAPERLESS_EXPORT_DIR /export
+RUN mkdir -p $PAPERLESS_EXPORT_DIR
+RUN chown -Rh paperless:paperless $PAPERLESS_EXPORT_DIR
+
 # Setup entrypoint
 COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
 RUN chmod 755 /sbin/docker-entrypoint.sh
 
 # Mount volumes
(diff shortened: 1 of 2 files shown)


danielquinn:
This looks ok to me, but my Docker-foo is still not very good.  @pitkley what do you think?

Dockerfile
```diff
@@ -35,12 +35,17 @@ RUN groupadd -g 1000 paperless \
     && useradd -u 1000 -g 1000 -d /usr/src/paperless paperless \
     && chown -Rh paperless:paperless /usr/src/paperless
 
+# Set export directory
+ENV PAPERLESS_EXPORT_DIR /export
+RUN mkdir -p $PAPERLESS_EXPORT_DIR
+RUN chown -Rh paperless:paperless $PAPERLESS_EXPORT_DIR
```
pitkley: Running this line during the build will not reliably work if the end-user host-mounts the export directory, this needs to be done in the `docker-entrypoint.sh` script. 

You can pretty much duplicate/copy-paste the logic from the [consumption directory](https://github.com/danielquinn/paperless/blob/819a0e1f570aa5e65d2bf3aa1b98e0bc13c19196/scripts/docker-entrypoint.sh#L18-L36).

Strubbl:
Thanks for the suggestion @pitkley . It would be nice if you can review my recent commit, too. I did not want to duplicate the code for the logic from the consumption directory, so i built a for loop.

scripts/docker-entrypoint.sh
```diff
@@ -15,26 +15,28 @@ map_uidgid() {
...
-        echo "failed."
-        echo ""
-        echo "Either try to set it on your host-mounted directory"
-        echo "directly, or make sure that the directory has \`o+x\`"
-        echo "permissions and the files in it at least \`o+r\`."
-    } >&2
-
+    # Set permissions for consumption and export directory
+    for i in PAPERLESS_CONSUMPTION_DIR PAPERLESS_EXPORT_DIR; do
+      cur_dir_name=$(sed -e's/PAPERLESS_//; s/_DIR//' <<< $i | tr '[:upper:]' '[:lower:]')
```
pitkley: While I wouldn't mind the message being "Changing group of PAPERLESS_CONSUMPTION_DIR", I appreciate the effort to get a normal sentence.

Three things (one I see as required, the other two optional):

1. *Required:* Please add a comment before the variable explaining what this does -- it took me a hot minute to understand what is going on. Maybe something like:

       # Extract the name of the current directory for the error message

2. *Optional:* I would change the loop-variable from `$i` to something more speaking, e.g. `$dir`. `$i` is usually used for integer variables.

3. *Optional:* I would favour using `awk` instead of both `sed` and `tr`, mainly because I think it enhances readability:

       cur_dir_name=$(echo "$i" | awk -F'_' '{ print tolower($2); }')

   But this is probably subjective, and with the added comment mentioned in (1), feel free to keep it as is! 👍

Strubbl:
Thanks for your feedback. I implemented all three suggestions. 
I am more familiar with sed, but your awk solution is much more readable.

danielquinn:
Great work on this guys, thanks so much!

PR #214, opened 2017-04-19
Fixes #213 (MySQL syntax error)

### LINKED ISSUE #213: ProgrammingError at /api/logs/ using MySQL
getting  the following error 

```
    (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL 
    server version for the right syntax to use near 'SEPARATOR '\n') AS `messages`, 
    MAX(`documents_log`.`modified`) AS `time` FROM `do' at line 1")
```

Could provide more info/versions upon request.

philippeowagner:


danielquinn:
As this change is pretty low-level, I'm going to have to test this manually on all three database types before I merge it.  If you were to include a unit test to make that easier, I'd appreciate it...

philippeowagner:
@danielquinn the change only affect MySQL, see the `if self.engine == self.ENGINE_MYSQL:` in line 52.

danielquinn:
Heh.  I didn't notice that line because it wasn't one of the ones you edited.

I'm pretty sure that I cribbed this code from an online howto somewhere, since I wouldn't have deliberately included support for MySQL from the beginning.  But if this is what it takes for it to work for you, then I'll merge it.

If it blow up for other MySQL users though, we'll have to see what the differences are, so do you mind posting here your server version just in case we need to go back and figure stuff out for others?

PR #219, opened 2017-04-24
paperless.conf.example: remove duplicate option

Strubbl:
This commit removes the duplicated option in this config:
please see https://github.com/danielquinn/paperless/blob/057d5f149fd63a34ebdd0fd00332e307ebcc3f95/paperless.conf.example#L113 compared with https://github.com/danielquinn/paperless/blob/057d5f149fd63a34ebdd0fd00332e307ebcc3f95/paperless.conf.example#L122


danielquinn:
Heh.  Nice catch.  Thanks :-)

PR #220, opened 2017-04-29
fuzzy matching

jgysland:
I occasionally get slightly erroneous OCR results that interfere with tagging in a fashion, and it's usually something like a slight misspelling or an extra or missing space which is generally difficult (or tedious) to account for using regular expressions. These kinds of errors are easily caught by a fuzzy string match, however, at the expense of an occasional false positive.

This is a very rudimentary implementation, but it's been useful to me, so I thought I'd share. There are several ways it could be improved, and I'll attempt to address feedback to get this merged or make it more useful.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ django-filter>=1.0
 django-flat-responsive>=1.2.0
 djangorestframework>=3.5.3
 filemagic>=1.6
+fuzzywuzzy[speedup]==0.15.0
 langdetect>=1.0.7
 pyocr>=0.4.6
 python-dateutil>=2.6.0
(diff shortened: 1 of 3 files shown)


danielquinn:
Actually, this is an excellent patch, complete with proper ordering of imports and a unit test.  The only request I have is that you remove the `(as implemented in the Python package \"FuzzyWuzzy\")` from the help_text since it's not really necessary that the user know the name of the underlying technology.

Other than that though, it's perfect!  Thanks so much for the patch.

jgysland:
Done. I just re-wrote explanation of fuzzy matching in that help text altogether; as it had been written it was somewhere between technical and simplified which isn't super useful.

danielquinn:
Excellent work.  Thank you!

PR #224, opened 2017-05-06
WIP : Exporter improvements

CkuT:
This is a work in progress in order to improve document import/export.

I'm open to improvements requests and reviews !

### DIFF
--- a/src/documents/management/commands/document_exporter.py
+++ b/src/documents/management/commands/document_exporter.py
@@ -62,17 +62,20 @@ class Command(Renderable, BaseCommand):
 
             document = document_map[document_dict["pk"]]
 
-            target = os.path.join(self.target, document.file_name)
-            thumbnail_target = target + "-tumbnail.png"
-            document_dict[EXPORTER_FILE_NAME] = target
-            document_dict[EXPORTER_THUMBNAIL_NAME] = thumbnail_target
+            file_target = os.path.join(self.target, document.file_name)
 
-            print("Exporting: {}".format(target))
+            thumbnail_name = document.file_name + "-tumbnail.png"
+            thumbnail_target = os.path.join(self.target, thumbnail_name)
+
+            document_dict[EXPORTER_FILE_NAME] = document.file_name
+            document_dict[EXPORTER_THUMBNAIL_NAME] = thumbnail_name
+
+            print("Exporting: {}".format(file_target))
 
             t = int(time.mktime(document.created.timetuple()))
-            with open(target, "wb") as f:
+            with open(file_target, "wb") as f:
                 f.write(GnuPG.decrypted(document.source_file))
-                os.utime(target, times=(t, t))
+                os.utime(file_target, times=(t, t))
 
             with open(thumbnail_target, "wb") as f:
                 f.write(GnuPG.decrypted(document.thumbnail_file))
--- a/src/documents/management/commands/document_exporter.py
+++ b/src/documents/management/commands/document_exporter.py
@@ -10,6 +10,7 @@ from documents.models import Document, Correspondent, Tag
 from paperless.db import GnuPG
 
 from ...mixins import Renderable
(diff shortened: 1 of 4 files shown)


src/documents/management/commands/document_exporter.py
```diff
@@ -71,6 +73,11 @@ def dump(self):
                 t = int(time.mktime(document.created.timetuple()))
                 os.utime(target, times=(t, t))
 
+            with open(thumbnail_target, "wb") as f:
+                f.write(GnuPG.decrypted(document.thumbnail_file))
+                t = int(time.mktime(document.created.timetuple()))
```
danielquinn: This line is redundant and could probably be moved just above line 71.  That way you can remove this line and line 73.

src/documents/management/commands/document_importer.py
```diff
@@ -77,7 +77,7 @@ def _check_manifest(self):
                 )
 
             doc_file = record["__exported_file_name__"]
-            if not os.path.exists(os.path.join(self.source, doc_file)):
+            if not os.path.exists(doc_file):
```
danielquinn: That's interesting.  Your change makes sense given the code in the exporter, but I'm not sure that the `__exported_file_name__` value is correct in the first place.  One should be able to export from one system with one directory structure to an entirely different system without things exploding, so if we're storing the full path in the export, that's a problem to me.

I can't test this right now as I'm out of town visiting family, but if you're tinkering in this area anyway, I'd like to know if you can do an export to one directory and an import from another one or if we have a problem in the manifest file here.

src/documents/management/commands/document_importer.py
```diff
@@ -91,9 +91,15 @@ def _import_files_from_manifest(self):
                 continue
 
             doc_file = record["__exported_file_name__"]
+            thumb_file = record["__exported_thumbnail_name__"]
             document = Document.objects.get(pk=record["pk"])
             with open(doc_file, "rb") as unencrypted:
```
danielquinn: nit: since we've now got two `with` blocks here, I'd appreciate it if you could add a new line around both as we have in a similar code block in the exporter.  It's just style really, but since you're here ... :-)

src/documents/management/commands/document_exporter.py
```diff
@@ -62,7 +62,9 @@ def dump(self):
             document = document_map[document_dict["pk"]]
 
             target = os.path.join(self.target, document.file_name)
+            thumbnail_target = target + "-tumbnail.png"
             document_dict["__exported_file_name__"] = target
+            document_dict["__exported_thumbnail_name__"] = thumbnail_target
```
danielquinn: Now that we've got two arbitrary strings being used to refer to stuff, they really should be re-defined as constants and imported both here and in the importer.  I recommend the following:

1. Create a file at `src/documents/settings.py`
2. Put the two constants in there (Or another reasonable name.  I'm not partial to these.):

```python
EXPORTER_FILE_NAME = "__exported_file_name__"
EXPORTER_THUMBNAIL_NAME = "__exported_thumbnail_name__"
```

3. In the importer and exporter use `from .settings import EXPORTER_FILE_NAME` (or whatever name you like)

danielquinn:
It all looks good and is clearly something that needs to be done for the import/export path, so thank you for doing the work.  My change requests are above, most of which are cosmetic.  The one troubling spot is the bit about the full path being in the name.  I can't test for this right now, but we may need to institute some sort of checking for full path vs. relative path somehow.  I'm not sure what we can do here yet, and won't know until I'm back home with a little more time to test this myself (about 2weeks).

CkuT:
Hey !

I made the changes you asked for, and I improved the export/import commands. It should be better now, with less errors.

I have tested an export-import with a complete different folder on my system and it seems to work. Can you test it on your side when you will have a little moment ?

danielquinn:
I've just managed to run it without issue locally, so thanks so much for your contribution!  I'll add this to the changelog in a few minutes.

PR #226, opened 2017-05-14
Bump pyocr requirement to version 0.4.7 to support tesseract 4.0.0alpha.

ddddavidmartin:
The [latest pyocr version](https://github.com/openpaperwork/pyocr/commit/85a011c4127b1faebd4890b9b423d692db5c3a08) now allows running pyocr with the latest tesseract version compiled from the tesseract master branch.

My oldish Ubuntu ships an ancient tesseract version, and being able to just compile and use the latest tesseract fills me with joy. Also, I hope that it produces better OCR results.

**Testing this**
On my setup it was simply a case of checking out the tesseract [master branch](https://github.com/tesseract-ocr/tesseract) and compiling + installing from there. It seems to work fine with paperless from the handful of documents I tried it with.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ djangorestframework>=3.5.3
 filemagic>=1.6
 fuzzywuzzy[speedup]==0.15.0
 langdetect>=1.0.7
-pyocr>=0.4.6
+pyocr>=0.4.7
 python-dateutil>=2.6.0
 python-dotenv>=0.6.2
 python-gnupg>=0.3.9


danielquinn:
This seems reasonable, but I'll have to test this out on my own before I merge it and that may be a few days.  Currently though, we're using `>=` and not `=`, so a default install should include the latest version anyway.

githubbrowser:
Hmmm, good point. I haven't actually tried what pip does without the change.
I wonder how pip handles the different cases when using `pip install --user --requirement /path/to/paperless/requirements.txt`.

* when pyocr 4.0.6 is installed, will it update? I would think not, because the requirement of >=4.0.6 is satisified
* when pyocr is not installed yet, will it actually grab the latest version?

Arguably requiring pyocr >=4.0.7 is only a change in the implementation version so it has little impact. But, it makes sure that in any case running paperless with a non-release tesseract works as expected.

danielquinn:
Alright I did a quick test of my own, and the latest version of pyocr (0.4.7) appears to work just fine.  While it's not necessarily a dependency for Paperless that we run the latest pyocr, I don't see any problem with bumping the version since the new version is backward compatible.  Merging now.

PR #227, opened 2017-05-21
Fix clened_data typos in forms.py.

ddddavidmartin:
I came across these typos whilst looking through the code. Not sure if and where these calls are used, as they would not work as they are. I have not tested this change but looked at it closely and deemed it right.


danielquinn:
Nice find!  Thanks for fixing it.

PR #228, opened 2017-05-24
Set email inbox in config file, fetch email at consumer startup and bring documentation up to date

ddddavidmartin:
I have been setting up the email consumption on my setup and noticed some bits where the docs are not current or trying out changes could be easier. Hence this pull request. Basically:

* Fetch emails at startup of the consumer so that after a config change it is immediately clear whether it works or not. I wondered whether this could be an issue if there is no internet connection yet at startup, but Paperless seems to handle that fine as it only prints an error but does not break.
* Allow to set the email inbox in the config file together with the other options.
* Bring the documentation of the email section up to date with the current implementation

Thanks!


danielquinn:
You know, you may be the only person using this mail feature other than me.  Thanks for these fixes.

PR #229, opened 2017-05-24
Copy Paperless service files to systemd directory before enabling them.

ddddavidmartin:
The problem with the original instructions is that systemd creates symlinks pointing to the service files in the paperless directory. A user is unlikely to leave the changes in the service files committed (especially not on a master branch checkout) and they are easily lost and the services fail to start without obvious reason.

To avoid this we simply copy the service files to the systemd directory directly and use the files in the repository only as an example.


danielquinn:
This is a much better system.  Thanks for your contribution here :-)

PR #230, opened 2017-05-26
Refer to Paperless in Django webserver titles and update Django documentation URLs

ddddavidmartin:
This is another simple one. Instead of referring to the Django website or the Django administration it looks much neater when it actually says Paperless. I'm far from an expert with Django so I have gone with the stackoverflow suggestion on how to set it. See screenshots below on how it looks.

Also, the weblinks in the documentation were still referring to version 0.9 which comes with a warning that it is unsecure. As per requirements we are actually on version 0.10 and the links in comments are updated accordingly.

screen shot 2017-05-27 at 8 38 51 am
screen shot 2017-05-27 at 8 39 23 am


danielquinn:
I've been doing Django stuff for almost ten years.  I had no idea this option existed.  Thanks for that!

PR #240, opened 2017-06-24
Add link to Django documentation for time zone setting in example config.

ddddavidmartin:
Trivial pull request but I was struggling with this twice already so I'm thinking it is good to include.

It is not obvious which time zones the option in the config file accepts. Having a link to the official django documentation makes it clear.

Thanks!


danielquinn:
Sure why not?  Thanks for the PR :-)

PR #250, opened 2017-07-31
create documents subfolder folder if they do not exist

brightdroid:
Create documents and subfolder in migration if they do not exist. This could happen if you start a new installation without creating these folders manually.


danielquinn:
Good catch, thanks for sharing :-)  I didn't know about `exist_ok=` either!

Issue #188, opened 2017-02-05
Paperless, but without GPG


guillaumededrie:
Hi,
@danielquinn, thank you very much for this useful project!

I just pushed a version of `Paperless` but without GPG. You can take a look at [the code](https://github.com/guillaumededrie/paperless-without-gpg).

I am not a really big fan of putting a GPG passphrase in a config file (mainly for security reason), so instead, I set up an encrypted directory on my NAS (key on RAM, so cold boot attack are possible) and Paperless do the rest! (I put a [readme](https://github.com/guillaumededrie/paperless-without-gpg) to explain in detail my setup).

At first it was just for me, but I realized that maybe some people can be interested by. 

I don't know if we can make an option (i.e. the user can choose not to encrypt PDFs), push the code on a different branch, or simply add a link somewhere in the README/Documentation.

Thank you again,

-- 
Guillaume

danielquinn:
Well the changes you've made there are rather drastic, so I honestly wouldn't consider supporting both methods -- the effort involved is just too much.  The project is GPL though, so if you wanna fork it for this purpose, go nuts :-)

A few suggestions though:

* Your changes include editing migrations which is a big no-no.  That's historical data for Paperless users, so making those sorts of changes would effectively make it impossible for someone to easily switch to an unencrypted setup like yours.  Either you start with an empty `migrations` directory, or you leave it alone and write a migration to decrypt your user's data so they can use your setup.
* Storing a password in a config file is definitely not optimal, but you'll note that [in the manage.py file there's an option to specify it at start time](https://github.com/danielquinn/paperless/blob/master/src/manage.py#L16) thereby saving you the dangers of storing the value in a config.

The truth is, I don't much like the use of GPG here, and in retrospect, I would have preferred to go with Python's excellent [cryptograpy](https://pypi.python.org/pypi/cryptography) module instead.  The real trouble spot however is that Paperless is expected not just to *store* sensitive data but to *serve it back up* on request, so whichever path I chose it had to be symmetric encryption unless we get into complex stuff like storing a private key in a cookie on each user's computer and... well that's a mess for a self-hosted project like this.

So to sum up, I'm glad you're finding Paperless so useful and that you were able to modify it for your own purposes.  Supporting a fork like yours though, which such drastic differences, isn't something I'm keen on doing.

Gummibeer:
I struggled over the same thing.
[...]

PR #251, opened 2017-08-07
Added option to disable encryption

mcronce:
I prefer to store my files in their raw form (already using full-disk encryption underneath), and stumbled across #188.

I noted your comment that the change was a bit heavy-handed - and after reviewing it I think that's putting it mildly :) - I think this will be more palatable.

Encryption stays enabled by default.  If the `DISABLE_ENCRYPTION` environment variable is set to `'true'`, then we (A) don't prompt for a passphrase when we find that it's empty on startup, and (B) just return the file's raw contents from `encrypted()` and `decrypted()` in `src/paperless/db.py`

Semantically speaking, I don't love doing it in those functions, but practically speaking, it seemed like the minimal possible change to achieve my goal.

### DIFF
--- a/src/manage.py
+++ b/src/manage.py
@@ -11,7 +11,7 @@ if __name__ == "__main__":
     # The runserver and consumer need to have access to the passphrase, so it
     # must be entered at start time to keep it safe.
     if "runserver" in sys.argv or "document_consumer" in sys.argv:
-        if not settings.PASSPHRASE:
+        if(settings.ENABLE_ENCRYPTION and not settings.PASSPHRASE):
             settings.PASSPHRASE = input(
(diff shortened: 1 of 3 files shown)


danielquinn:
Unfortunately, your suggestion isn't enough to roll into master.  It's just too easy for people to accidentally corrupt their document store with some encrypted and some unencrypted files.  I did however take your changes and roll them into [a new branch](https://github.com/danielquinn/paperless/pull/254/files) (WIP) that might give you what you need while also safeguarding other users.  It's not ready yet as I have to include documentation and handle the importer & exporter, but it's a start.

I'm getting married this month though, so progress on this is gonna be **slow**.

danielquinn:
I'm going to close this now as my WIP branch has it's own PR at #254.

PR #254, opened 2017-08-12
WIP: Allow encryption to be disabled

danielquinn:


matthewmoto:
I have this patch-set working in my fork and noticed your CI build is failing because you're missing the migration file from running "./manage makemigrations" to get Django to pick up the schema change for the new encryption field. Are you actively working on this or would you like me to throw together a cleaned-up pull-request for integration?

danielquinn:
Hi @matthewmoto and thanks for reminding me that I still have this on the backburner.  Yeah I hadn't committed that file yet 'cause I wasn't sure I was happy with the changes.  I'll give this another look today and rebase it from `master` as a lot has changed since my last commit.  I'll at least get it to a point where the tests are passing, and if I have the energy, we'll see about merging it.  I just have to make sure this isn't going to break other people's setups first.

danielquinn:
Ok I've reworked the patchset to be a little smarter and future-proof, so if you don't mind @matthewmoto, could you give it a try when you've got a minute?

Specifically I've dropped the use of a `DISABLE_ENCRYPTION` boolean in favour of a more specific `STORAGE_TYPE` value.  This will allow me to change the storage types in the future to include stuff like `public-key` or even `s3`.

The tests all work for me, but I'd like a second pair of eyes on this if you've got the time.

matthewmoto:
Thanks, @danielquinn for the quick turnaround. I just pulled these new changes and tested them with a handful of use cases to get away from the happy-path:
- importing with encryption on and off and flipping back and forth to make sure everything is readable afterward
- exercising some of the newer changes like the pdf with pre-existing text to ensure non-obvious badness doesn't happen with this (since I haven't pored over those changes with much detail yet).
- checking the startup logic with the passphrase unset and set to make sure it complains appropriately.

I like the way you changed the storage type to something that opens things up in the future for more back-ends. It also implies a future where documents can have different storage types (a cool future to think about with per-user passwords or simply non-local files).

I did find two things that consider issues (not that they should hold-up this merge). I wanted to pass them by you here before clogging up the issue tracker to see if you think they're worth noting:
 1. In testing with the passphrase un-set, but gpg enabled, I noticed that the startup complaint logic lets you simply hit enter twice and it will let you proceed with an empty passphrase and start the server anyway. I kinda stumbled into it accidentally since I fat-fingered the return key. It might be worth changing things a bit like this (or with a retry-count):
```
--- manage.py 2018-02-05 08:07:39.356602800 -0500
+++ manage.proposed.py  2018-02-05 08:43:43.201131300 -0500
@@ -12,7 +12,7 @@
     # must be entered at start time to keep it safe.
     if "runserver" in sys.argv or "document_consumer" in sys.argv:
         if not settings.STORAGE_TYPE == "unencrypted":
-            if not settings.PASSPHRASE:
+            while not settings.PASSPHRASE:
                 settings.PASSPHRASE = input(
                     "settings.PASSPHRASE is unset.  Input passphrase: "
                 )
```

 2. The way we currently handle changed passphrases doesn't fail gracefully. I realize there is a large friendly "DON'T CHANGE PASSPHRASES" in the config file, but somebody at some point is going to do it (I did inadvertently with the case mentioned above). In doing so, I noticed the thumbnails are broken and file downloads **succeed**, but empty contents with no error logs from the server.  

A simple solution to this could be to save the hash of the current password somewhere in the database so it can be compared at startup time to prevent these problems (or prompt to launch automatic migration behavior). Or, the very least, logging some errors to tell the user what actually went wrong (even if it's just some server logs to the console) may save some users their sanity.

danielquinn:
These are both good suggestions.  Sorry it's taking so long to get to this, but as you might have noticed, I'm juggling a lot with this project :-)

I'll try to implement your suggestions in the coming week(s).

matthewmoto:
No need to apologize. The only reason I just mentioned it instead of proposing it as a pull request is that I'm in the same boat with the juggling (life, job, etc). I'll also try to make some time to put together some patches to contribute in the next week or so.  It takes a village.

danielquinn:
Well it's been 27days, but I've just added some changes that should address both of your points above.

Notably, rather than storing a hash of the old password (which wouldn't detect switching back and forth and would require a new db column), I chose to go the route of checking if the first document in the db was stored with the same password we're currently using.  It's not a 100% perfect check though.  For that, we'd have to check *all* stored files, but I tried that, and start up was *really* slow.  The way I figure it though, at least this way, the user will get a warning the first time they change the password.  If they then store a few documents anyway and then switch the password back, that's on them.

@mcronce can you give this a try when you've got a minute?  If it pans out for you, I'll merge it for 1.4.0 or 1.5.0.

BastianPoe:
Is there anything I can do to get this PR closer to being merged? I need this feature. ;-)

danielquinn:
@BastianPoe if you can give it a test and let me know if it works for you, I'll feel more comfortable merging it.  Ideally I'd like to know how it works for you when you try to switch to the unencrypted mode, and if the switching path works for you.

GarethFox:
Hello fellows !
I would be also interested in beeing able to disable the encryption storage of processed documents.
How can we enjoy the feature on a docker-compose image startup ?
Thanks for all,

erikarvstedt:
Patching Paperless to remove the encryption was the first thing I did when I started using it.

I'd strongly vote for disabling encryption by default (or for removing it entirely) because it lends users a false sense of security unless they make the conscious effort to also store the database on an encrypted medium. (The database contains the unencrypted text contents of all encrypted docs.)
And if users have an encrypted medium, they can use it to store the docs and don't need the builtin encryption.
What's the use case you had in mind with encryption for docs only?

I've just tested this PR.
The `changed_password_check` introduced in https://github.com/danielquinn/paperless/pull/254/commits/e895a92f05c753f44b30de0230aff18cd29cbdc6 causes an error when running `manage.py migrate` on an empty DB, because the check happens before the required tables are created.

Besides this, everything works great! :+1:  
I've [rebased this PR](https://github.com/erikarvstedt/paperless/commits/ea-disable-encryption) to the current master (I [fixed](https://github.com/erikarvstedt/paperless/commit/0a1c4fae8bad27b96f07bf00be313bbef17fa012) the colliding migration numbers) and went through all possible usage paths, including a migration from data with encrypted docs.

GarethFox:
Hi @erikarvstedt,
My use cases depend on the related projects, some would not require any encryption (pdf files + ocr text in db), some other would require full encryption (pdf files + ocr text in db)
Unfortunately I am not python friendly, would you mind explaining how it is possibe to disable the gpg encryption of the processed document and keep them moved in *.pdf basic format once the consumer is done with it?

erikarvstedt:
@GarethFox, to keep the discussion on-topic, I've replied to you [here](https://github.com/erikarvstedt/unencrypted-paperless/issues/1). If you have any further questions, please post them in that other thread.

danielquinn:
Ok it looks like this PR is ready then, but @erikarvstedt I'm afraid I don't know how to make use of your fixes where you've rebased this PR.  Did you want to issue a PR of your own based on those changes, or is there something I can do to pull your changes into this PR?  I've never done something like that.

erikarvstedt:
My rebase fixes are trivial: I've just renamed the conflicting migration). You can copy them without attribution.
But please be aware that there's still a bug in your PR, as noted above.
In my rebase I just omitted the faulty commit, but I'm sure you'd want to include it.

danielquinn:
Alright it's getting late in the day here so I think I'm done tinkering with Paperless for today.  This is next on the list though, for when I have more time.

erikarvstedt:
By the way, I still think encryption should be disabled by default.
I'd love to hear your thoughts on my [objections above](https://github.com/danielquinn/paperless/pull/254#issuecomment-387103504).

danielquinn:
When I first started Paperless, I added encryption pretty early on because the documents themselves (not just their content) were sensitive.  Things like high-resolution scans of my passport, and full-colour copies of other government documents.  It's one thing to have my passport number, but to have a quality *copy* of it is another.

With that said though, the encryption *isn't very good* as it relies on a symmetric algorithm, meaning if you have access to the server (and the user running Paperless) you can read the files.  Still, for people who might be storing the `media` directory on another server like S3 or whatever, this is a nice extra layer of security.

In a perfect world, I'd want async encryption, with a step that requires that the user copy/paste their private key into the server at runtime (no config file).  That's a pain in the ass for most people though, so I don't think that'll happen unless someone has a better idea for making this more user-friendly.

Turning encryption off by default is a good idea, if only because it's not immediately obvious to people when they start playing with Paperless that that's what it's doing.  However, given that all we have to work with are environment variables, turning it off would mean a lot of people *thinking* their stuff was encrypted when it wasn't.

Your suggestion that we only turn it off for people who don't already have an encrypted file introduces a degree of unpredictability that I'm not comfortable with.  In such a system, people can have fully encrypted libraries with encryption turned off in their config files.  If those people then spin up another instance of Paperless with the same config, they'll have different results.

I think that a better solution might be to follow these rules:

* **If** encryption is enabled in the config, use it
* **If** encryption is disabled in the config, don't use it
* **If** encryption isn't set **and** there's no encrypted files, don't use it
* **If** encryption isn't set **and** there's even one encrypted file, trigger an error that asks that they explicitly state what they want.

erikarvstedt:
I agree with the bullet points, especially the last one is a great idea!

What do you think of the following?
- Drop the `PAPERLESS_STORAGE_TYPE` setting and let `PAPERLESS_PASSPHRASE` imply that the database is encrypted.
- Keep the document field `storage_type`. <sup>[1]</sup>
- On startup: If one doc has `storage_stype == encrypted`, demand that `PAPERLESS_PASSPHRASE` is set.

What we lose:
- The ability to add unencrypted docs to a database that already includes encrypted docs.
  I don't see how this could even be useful. Users would gain more from a guide in the documention that explains how to unencrypt an existing db by exporting and re-importing.

What we gain:
- Conceptual clarity and a simplified implementation.
- All the bullet points are satisfied.
- No expensive `changed_password_check` at startup for encrypted databases.
  (`GnuPG.decrypted` is called for every doc file).

I'd happily offer to implement all of the above.



**[1]** We could also use a boolean document field `encrypted` or a single global boolean variable for the whole DB if this possible with Django.

danielquinn:
Ok, here's an update on this Big Ticket:

* Encryption is now **off** by default.
* I've taken @erikarvstedt's advice on dropping `STORAGE_TYPE` in favour of just using the existence of `PAPERLESS_PASSPHRASE`.  Encryption will only be used if `PAPERLESS_PASSPHRASE` is set to something.
* I've also added a management command that people can use to convert their entire document store from an encrypted state an unencrypted one (or vice-versa).
* As this is a major change, I figure semver dictates the version number be bumped to 2.0.0.
* The config files are documented, but the official documentation hasn't been patched yet.
* The tests (what few there are anyway) all pass, and I've managed to run the migrations against my own database *and* a brand new installation.
* Finally, I ran the encryptor/decryptor script against my own database and there weren't any problems.

Guys, please have a look at this and test this branch against your own stuff if you can.  Notably, the Docker stuff isn't something I usually tinker with, so it'd be nice if there was a voice in here saying that it works for them in their Docker setup.

Outside of that, I'm going to spend some more time fixing the documentation up to include the new option to keep things unencrypted etc., after which I expect to merge this unless Something Terrible is discovered.

erikarvstedt:
Checked all usage and migration scenarios, the conversion command and the docker setup.
Looks great overall! :+1: 
I've added a [few suggestions](https://github.com/erikarvstedt/paperless/commits/disable-encryption) (see the commit messages for details).
Only [this commit](https://github.com/erikarvstedt/paperless/commit/dc80c32298adcf5960eb5025a893bd923b7b14bd) contains important user-facing changes.
Feel free to squash any of these commits into your original commits, no need to mention my name.

This PR needs a rebase or a merge to include the conflicting changes in master.
Will test again when this is done.

erikarvstedt:
I've [merged this PR](https://github.com/erikarvstedt/paperless/commits/disable-encryption-merged) and re-tested it extensively, with migrations from the current master (https://github.com/danielquinn/paperless/commit/d5876cc97d740856171dca8afb21370d8421e5e7).

Is there anything else I could do to help getting this released?

erikarvstedt:
In your merge (https://github.com/danielquinn/paperless/pull/254/commits/c9f35a7da2cb1cb99754b2c1d16592fffbcdb784) you accidentally re-introduced the [following lines](https://github.com/danielquinn/paperless/blob/c9f35a7da2cb1cb99754b2c1d16592fffbcdb784/src/documents/consumer.py#L57) which are obsolete: 
```
self.stats = {}
self._ignore = []
```
Do you have any objections to just using my merge commit? It's all thoroughly tested and the history is less convoluted.
If you need any help with git, just let me know.

danielquinn:
Sorry @erikarvstedt, my git-foo scores about a 5/10 -- enough to do the job, but not enough to feel secure in taking one branch, merging another one from another repo, checking the diff, and knowing that everything is ok to merge into master.  As you made it sound easy to catch the differences, I opted just to do the merge and fix the conflicts myself instead of trying to figure out the above.

Did you do anything other than fix the conflict & the migration conflict?

danielquinn:
I just had a look at your parallel branch to see what might be different.  A few things are good to merge, but others I'm not cool with.  For example, I want to keep `.storage_type` because I want to allow for future storage engines (public key encryption, S3, etc.)  Similarly, `change_storage_type` makes more sense than `change_encryption` if you want to allow for the above.

The way you joined the migrations won't play well for Django though, as it's likely that there are other users out there who have already run migration 20, adding stuff to it will mean that those users will skip that part of the migration.

I can remove the emoji from the output of the storage_type changer if you like.  I thought it was cute, but if it doesn't work for you, you're probably not alone.  As for the other two commits: fixing the docstring and improving the password check, can you tell me how to cherry-pick those into this branch?  Either that, or just wait 'til this is merged and then you can issue a separate pull request.

erikarvstedt:
No, just fixed conflicts the in my merge commit.

Your tweaks of the export command are great! But [my suggested commits](https://github.com/erikarvstedt/paperless/commits/disable-encryption) will need a little update now. If you like them, I'll rebase them to your recent changes.

erikarvstedt:
To cherry-pick the [password check commit](https://github.com/erikarvstedt/paperless/commit/dc80c32298adcf5960eb5025a893bd923b7b14bd), just run
```
curl -L https://github.com/erikarvstedt/paperless/commit/dc80c32298adcf5960eb5025a893bd923b7b14bd.patch | git am
```

danielquinn:
Oh I see what you mean about the migrations now.  I'll have a look when I get back from dinner.  I'll be merging this tonight though!

erikarvstedt:
Would you be okay with using an integer for storing the storage type, [like so](https://github.com/erikarvstedt/paperless/commit/280e962254806ce1b8c68ce93cd69ff48a2ba803)?
I feel a bit uneasy with the needless string comparisons and the db bloat.

Enjoy your dinner! :pizza:

danielquinn:
TIL something new about git, thank you! :-D

Alright, I've cherry-picked two of your commits, but I don't think I'm ok with switching `storage_type` to an integer field.  I used to do things that way, but I found that it lead to a database that was just harder to understand when I'm operating on it at the SQL level.  At least with strings like `gpg` and `unencrypted`, it's obvious what's going on when you do a `SELECT * FROM documents;`.

I'm merging now!

PR #255, opened 2017-08-19
add FORCE_SCRIPT_NAME to host paperless on a subpath url

maphy-psd:
I host paperless on raspberrypi/paperless. This should do the trick.

### DIFF
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -71,12 +71,12 @@ class DocumentAdmin(CommonAdmin):
 
     def thumbnail(self, obj):
         if settings.FORCE_SCRIPT_NAME:
-            _src="{}/fetch/thumb/{}".format(settings.FORCE_SCRIPT_NAME,obj.id)
+            src_link = "{}/fetch/thumb/{}".format(settings.FORCE_SCRIPT_NAME, obj.id)
         else:
-            _src="/fetch/thumb/{}".format(obj.id)
+            src_link = "/fetch/thumb/{}".format(obj.id)
         png_img = self._html_tag(
             "img",
-            src=_src,
+            src=src_link,
             width=180,
             alt="Thumbnail of {}".format(obj.file_name),
             title=obj.file_name
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -70,9 +70,13 @@ class DocumentAdmin(CommonAdmin):
     created_.short_description = "Created"
 
     def thumbnail(self, obj):
+        if settings.FORCE_SCRIPT_NAME:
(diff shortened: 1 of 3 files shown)


src/documents/admin.py
```diff
@@ -70,9 +70,13 @@ def created_(self, obj):
     created_.short_description = "Created"
 
     def thumbnail(self, obj):
+        if settings.FORCE_SCRIPT_NAME:
+            _src="{}/fetch/thumb/{}".format(settings.FORCE_SCRIPT_NAME,obj.id)
```
MasterofJOKers: There a space missing after the `,`.

src/paperless/settings.py
```diff
@@ -47,7 +47,11 @@
 if _allowed_hosts:
     ALLOWED_HOSTS = _allowed_hosts.split(",")
 
-
+FORCE_SCRIPT_NAME = None
+_force_script_name = os.getenv("PAPERLESS_FORCE_SCRIPT_NAME")
+if _force_script_name:
+    FORCE_SCRIPT_NAME = _force_script_name
```
MasterofJOKers: Looking at https://docs.python.org/3/library/os.html#os.getenv, you could just write `FORCE_SCRIPT_NAME = os.getenv("PAPERLESS_FORCE_SCRIPT_NAME")` as this would return `None` per default. There's no need to set it to `None` explicitly.

src/documents/admin.py
```diff
@@ -70,9 +70,13 @@ def created_(self, obj):
...
+        else:
+            _src="/fetch/thumb/{}".format(obj.id)
         png_img = self._html_tag(
             "img",
-            src="/fetch/thumb/{}".format(obj.id),
+            src=_src,
```
danielquinn: Please don't prefix local variables with an underscore like this.  It makes sense for instance variables or class methods (as the underscore denotes privacy), but in a local variable it means nothing.

src/documents/admin.py
maphy-psd: Thanks..i will fix it.

src/paperless/settings.py
maphy-psd: Oh thanks..i still have to learn a lot

src/documents/admin.py
maphy-psd: Sry..you're right. Is src_link ok?

danielquinn:
Thanks for contributing this.  I've been (and still am) on my honeymoon, hence the delay in merging it.

The changes look good, and I appreciate your work to conform to pep8 etc.  I'll merge it now, and then add some documentation later this week before doing a version bump.

PR #256, opened 2017-08-26
Add financial year documents filter

ddddavidmartin:
I'm using Paperless to store my tax documents and being able to filter by financial year is quite handy. So, this pull request brings in support for a financial year filter.

The start and end date must be set in the config but everything else pretty much just works. Both financial years that go from Jan-Dec as well as for example Jul-Jun work.

screen shot 2017-08-26 at 8 05 03 pm

Thanks!

### DIFF
--- a/paperless.conf.example
+++ b/paperless.conf.example
@@ -162,6 +162,12 @@ PAPERLESS_PASSPHRASE="secret"
 #PAPERLESS_TIME_ZONE=UTC
 
 
+# If set, Paperless will show document filters per financial year.
+# The dates must be in the format "mm-dd", for example "07-15" for July 15.
+#PAPERLESS_FINANCIAL_YEAR_START="mm-dd"
+#PAPERLESS_FINANCIAL_YEAR_END="mm-dd"
+
+
 # The number of items on each page in the web UI.  This value must be a
 # positive integer, but if you don't define one in paperless.conf, a default of
 # 100 will be used.
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -63,9 +63,9 @@ class FinancialYearFilter(admin.SimpleListFilter):
 
     def _determine_fy(self, date):
         """Return a (query, display) financial year tuple of the given date."""
-        fy_start = self._fy_start(date.year)
-
         if self._fy_does_wrap():
+            fy_start = self._fy_start(date.year)
+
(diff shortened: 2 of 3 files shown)


ddddavidmartin:
I haven't looked at implementing tests for this yet, mostly as I'm not quite sure where to begin I reckon.

thomasbrueggemann:
I also use paperless for managing my financial documents, but use tags for this. 
I assign those documents tags such as "tax-2017-Q1".
Might be a usecase for you too, instead of baking specialty functionality into the otherwise quite generic paperless? Just an idea.

ddddavidmartin:
Hmmm, you are making a good point there @thomasbrueggemann. You can't do quarters easily with that filter for example.

My thought process:
* Paperless does not support applying tags for a range of dates, it only matches by content. As a consequence I would have to tag all my documents by hand, right?
* A financial year is just a range of days in a year, and all documents by default have a date. So you get that categorisation into the specific financial years basically for free. I do use tags to mark documents for taxes and specific projects in general though.
* The financial year filter is not active unless it is set in the config.

ddddavidmartin:
One thing that bugs me about the filters is that they iterate through _all_ documents when you open the overview to generate the tables. That just seems wrong but I'm not sure how to avoid it without diving way deeper into it. Tags are much more efficient that way. :)

danielquinn:
Anything that lets users better find their docs sounds good in my book and this code is very clean, so I'm inclined to merge it, but you two seem to be unsure about it.

While I like @thomasbrueggemann's idea of just using tag names, I'm not sure it's the best place to track this kind of data unless it's reasonable to say that some documents might contain a date *outside* a specific financial year whilst still *belonging* to that year.  In that case, the date on the document can't be trusted, so a tag makes sense and this PR doesn't.  However, if that's very unlikely, then loading this functionality into tags doesn't make sense since the date alone gives us everything we need and it's already there.

So as I'm not one to think much about the concept of a financial year, I'm looking for some advice here.  What do you guys think?

danielquinn:
Also, hello from Ιος!

ios

ddddavidmartin:
I'm a little bit envious of your views but I'm happy for you to merge it in @danielquinn. :)

It may not be a one-fits-all perfect solution as per the discussion but achieving the same with tags looks like more work in the end.

PR #257, opened 2017-09-11
PAPERLESS_OCR_LANGUAGES -> PAPERLESS_OCR_LANGUAGE

caffeineflo:
I've seen discrepancies in the documentation regarding the use of both `PAPERLESS_OCR_LANGUAGES` and `PAPERLESS_OCR_LANGUAGE` (note singular/plural) and looking at the source I came to believe only the singular version is being used.

What I did is, I changed the plural versions to the singular version to avoid confusion and stay conform to what is being evaluated in code.

Please correct me if I'm wrong here

### DIFF
--- a/docker-compose.env.example
+++ b/docker-compose.env.example
@@ -8,7 +8,7 @@ PAPERLESS_PASSPHRASE=CHANGE_ME
 # PAPERLESS_OCR_THREADS=4
 
 # Additional languages to install for text recognition
-# PAPERLESS_OCR_LANGUAGES=deu ita
+# PAPERLESS_OCR_LANGUAGE=deu ita
 
 # You can change the default user and group id to a custom one
 # USERMAP_UID=1000
--- a/docker-compose.yml.example
+++ b/docker-compose.yml.example
@@ -16,7 +16,7 @@ services:
         # languages the user might have set in the env-file by overwriting the
         # value with nothing.
         environment:
-            - PAPERLESS_OCR_LANGUAGES=
+            - PAPERLESS_OCR_LANGUAGE=
         command: ["runserver", "0.0.0.0:8000"]
 
     consumer:
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -187,7 +187,7 @@ Docker Method
      document pages in parallel. If the variable is not set, Python determines
      the core-count of your CPU and uses that value.
 
-   ``PAPERLESS_OCR_LANGUAGES``
+   ``PAPERLESS_OCR_LANGUAGE``
      If you want the OCR to recognize other languages in addition to the
      default English, set this parameter to a space separated list of
      three-letter language-codes after `ISO 639-2/T`_. For a list of available
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -82,8 +82,8 @@ if [[ "$1" != "/"* ]]; then
     initialize
 
     # Install additional languages if specified
-    if [ ! -z "$PAPERLESS_OCR_LANGUAGES"  ]; then
-        install_languages "$PAPERLESS_OCR_LANGUAGES"
+    if [ ! -z "$PAPERLESS_OCR_LANGUAGE"  ]; then
+        install_languages "$PAPERLESS_OCR_LANGUAGE"
     fi
 
     exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"


ddddavidmartin:
I haven't worked with the Docker side at all, but to me it looks like it should be `PAPERLESS_OCR_LANGUAGES` across the board rather than `PAPERLESS_OCR_LANGUAGE`. The setting determines additional language**s** to be used rather than just a singular one. No?

danielquinn:
I don't use the Docker system either and I usually run this sort of thing by @pitkley since he hammered together the Docker setup we have.  What say you Pit?

caffeineflo:
I'm not using the docker stuff either, it was just one of the places where the singular form occurred as well.

I found this initially when reading through the docs and the example config file, as the docs are using the plural form while the example config uses the singular form.

I'm still unsure which is evaluated but looking at the source it seemed like the singular form is being used.

pitkley:
`PAPERLESS_OCR_LANGUAGE` (singular) is used to define the default language used by Tesseract when running the first OCR (see this chain of assignments: [1], [2], [3]). You should probably set this to the language you are importing into Paperless most often, since the first pass will be the correct one then.

`PAPERLESS_OCR_LANGUAGES` (plural) is only used in the Docker context to allow you to install additional OCR languages into the what would otherwise be a static Docker image that only contains english OCR. The environment variable gets [passed to `install_languages`][4] which will then try to [install every requested language][5].

Replacing the plural by the singular will break the default language if you want to install more than one additional language into the Docker image.

[1]: https://github.com/danielquinn/paperless/blob/2c55aad6c0e3da7122ac56c51119d34240900f88/src/paperless/settings.py#L208
[2]: https://github.com/danielquinn/paperless/blob/b7cb70805379445815b0d0a76c3c315e3bb492b9/src/paperless_tesseract/parsers.py#L33
[3]: https://github.com/danielquinn/paperless/blob/b7cb70805379445815b0d0a76c3c315e3bb492b9/src/paperless_tesseract/parsers.py#L114
[4]: https://github.com/danielquinn/paperless/blob/c5632e5c04a901c65f5bf9dea80a5de0350818b4/scripts/docker-entrypoint.sh#L86
[5]: https://github.com/danielquinn/paperless/blob/c5632e5c04a901c65f5bf9dea80a5de0350818b4/scripts/docker-entrypoint.sh#L63-L74

danielquinn:
@pitkley, what do you think about dropping `PAPERLESS_OCR_LANGUAGES` altogether and just installing all language packs by default?  Is the size differential really that big, or is there another reason someone might not want to have all languages available?

pitkley:
I see a two big disadvantages to this:

1. Image size

    Installing all currently available languages:

    ```
    The following NEW packages will be installed:
      libgif4 liblept4 libopenjp2-7 libtesseract-dev libtesseract3 tesseract-ocr tesseract-ocr-afr tesseract-ocr-ara
      tesseract-ocr-aze tesseract-ocr-bel tesseract-ocr-ben tesseract-ocr-bul tesseract-ocr-cat tesseract-ocr-ces
      tesseract-ocr-chi-sim tesseract-ocr-chi-tra tesseract-ocr-chr tesseract-ocr-dan tesseract-ocr-deu tesseract-ocr-deu-frak
      tesseract-ocr-dev tesseract-ocr-ell tesseract-ocr-eng tesseract-ocr-enm tesseract-ocr-epo tesseract-ocr-equ
      tesseract-ocr-est tesseract-ocr-eus tesseract-ocr-fin tesseract-ocr-fra tesseract-ocr-frk tesseract-ocr-frm
      tesseract-ocr-glg tesseract-ocr-grc tesseract-ocr-heb tesseract-ocr-hin tesseract-ocr-hrv tesseract-ocr-hun
      tesseract-ocr-ind tesseract-ocr-isl tesseract-ocr-ita tesseract-ocr-ita-old tesseract-ocr-jpn tesseract-ocr-kan
      tesseract-ocr-kor tesseract-ocr-lav tesseract-ocr-lit tesseract-ocr-mal tesseract-ocr-mkd tesseract-ocr-mlt
      tesseract-ocr-msa tesseract-ocr-nld tesseract-ocr-nor tesseract-ocr-osd tesseract-ocr-pol tesseract-ocr-por
      tesseract-ocr-ron tesseract-ocr-rus tesseract-ocr-slk tesseract-ocr-slk-frak tesseract-ocr-slv tesseract-ocr-spa
      tesseract-ocr-spa-old tesseract-ocr-sqi tesseract-ocr-srp tesseract-ocr-swa tesseract-ocr-swe tesseract-ocr-tam
      tesseract-ocr-tel tesseract-ocr-tgl tesseract-ocr-tha tesseract-ocr-tur tesseract-ocr-ukr tesseract-ocr-vie
    0 upgraded, 74 newly installed, 0 to remove and 0 not upgraded.
    Need to get 254 MB of archives.
    After this operation, 688 MB of additional disk space will be used.
    ```

    So we would gain at least 254 MB, almost doubling the size of the current image from 351 MB (the Python base image is 273 MB of that).

2. Tesseract language detection

    I might be wrong here, but I would assume that in the Tesseract language detection phase, Tesseract checks whatever it has OCRed with all installed languages to determine the best match. This might increase processing time, potentially a lot, while the regular end-user probably only needs two languages to be checked.

I think better documenting the difference between `PAPERLESS_OCR_LANGUAGE{,S}` rather than removing it would be the better move.

caffeineflo:
To add one more thing to the discussion:

Adding all languages probably won't be nice to the docker image, I agree with @pitkley on this.
Yet, I think there's still something unclear on this topic. Here's what I found and what's my use case:
I have both documents in German and in English, therefore I installed the language packages for deu and eng. 
Setting `PAPERLESS_OCR_LANGUAGE` to `eng deu` as the Dockerfile suggest is one way to do it (I know what it's actually used for, but as the documentation seems to be unclear on this, one might could think about it this way)  makes the paperless-consumer crash:

`pyocr.error.TesseractError: (1, b'Error opening data file /usr/share/tessdata/deu eng.traineddata\nPlease make sure the TESSDATA_PREFIX environment`

Naming the parameter `PAPERLESS_OCR_LANGUAGES` and setting eng deu works, but from looking at the source I'm almost sure it is being ignored from paperless itself. 
For whatever reason, my documents are still being ocr'ed for both languages, but that would have to be dependent on my installed packages then, right?! (Which also shouldn't be the case I guess)

Sorry for causing all this trouble, just trying to help improve either the documentation or the source files.

pitkley:
The way it should work right now is this:

* `PAPERLESS_OCR_LANGUAGES` (plural) is only used as part of the Docker image to install additional required languages at runtime to get around installing all of them at build time. Otherwise this variable is not used anywhere else, at least it shouldn't be.

* `PAPERLESS_OCR_LANGUAGE` (singular) is used by Paperless itself, the Docker image/runtime does not use it at all. This variable is used to determine the default language to use, that is: the language to OCR with first.

    The OCR process at a top level currently works like this:

    1. Extract images out of a given PDF
    2. Use Tesseract to OCR the images using a default language, as given by `PAPERLESS_OCR_LANGUAGE`
    3. Use Tesseract to guess the language based on the OCRed text from step 2
    4. If the guessed language does not equal the default language used in step 2, run OCR again with the guessed language

    For example, if you are scanning both German and English documents, with most of the documents being German, setting `PAPERLESS_OCR_LANGUAGE=deu` will speed up the OCR process for your German documents, since it will be done after step 3 and will not require an additional OCR run.

You say your documents are being OCRed for both languages. Can you expand on that: what are `PAPERLESS_OCR_LANGUAGE` and `PAPERLESS_OCR_LANGUAGES` set to, and what language are the documents you are seeing this behaviour with?

danielquinn:
It sounds like the consensus here is that this shouldn't be adopted.  This, coupled with the fact that the branch has had conflicts for a while now, is enough for me to close it.

Issue #267, opened 2017-10-30
Docker: wrong folder permissions for /consume


dev-rke:
Hi @danielquinn, hi @pitkley,

When starting up the application via docker-compose i get the following warning:
```
chgrp: cannot access ‘’: No such file or directory
Changing group of export directory:
  
failed.
2017-10-30T13:40:13.134346296Z 
Either try to set it on your host-mounted directory
directly, or make sure that the directory has `o+x`
permissions and the files in it at least `o+r`.
chmod: cannot access ‘’: No such file or directory
Changing group permissions of export directory:
  
failed.
2017-10-30T13:40:13.134371036Z 
Either try to set it on your host-mounted directory
directly, or make sure that the directory has `o+x`
permissions and the files in it at least `o+r`.
```

This seems to lead to the following issue, that the consumer dies when files cannot be removed after parsing:
```
Starting document consumer at /consume
Parsers available: RasterisedDocumentParser
Consuming /consume/image.jpg
Processing sheet #1: /tmp/paperless/paperless6vul_yxo/convert-0000.pnm -> /tmp/paperless/paperless6vul_yxo/convert-0000.unpaper.pnm
OCRing the document
Parsing for eng
Completed
Traceback (most recent call last):
  File "/usr/src/paperless/src/manage.py", line 18, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/usr/src/paperless/src/documents/management/commands/document_consumer.py", line 57, in handle
    self.loop()
  File "/usr/src/paperless/src/documents/management/commands/document_consumer.py", line 67, in loop
    self.file_consumer.consume()
  File "/usr/src/paperless/src/documents/consumer.py", line 139, in consume
    self._cleanup_doc(doc)
  File "/usr/src/paperless/src/documents/consumer.py", line 223, in _cleanup_doc
    os.unlink(doc)
[...]

PR #268, opened 2017-10-31
Set `g+w` on the consumption/export directories

pitkley:
This should fix issue #267.


danielquinn:
I just wanna say how fab it is that you've supported the Docker component for this for as long as you have.  Thanks :-)

PR #274, opened 2017-12-09
adapted Dockerfile for alpine image

addadi:
moving to alpine based image saves appx. 400MB (45%) of the docker image size, with same functionality

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,54 +1,41 @@
-FROM alpine:latest
-
+FROM alpine:3.7
+WORKDIR /usr/src/paperless
+COPY requirements.txt /usr/src/paperless/
+# Copy application
+COPY src/ /usr/src/paperless/src/
+COPY data/ /usr/src/paperless/data/
+COPY media/ /usr/src/paperless/media/
+# Set export directory
+ENV PAPERLESS_EXPORT_DIR /export
+# Set consumption directory
+ENV PAPERLESS_CONSUMPTION_DIR /consume
+COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
 # Install dependencies
 RUN apk --no-cache --update add \
         python3 gnupg libmagic bash \
         sudo tesseract-ocr imagemagick ghostscript unpaper && \
     apk --no-cache add --virtual .build-dependencies \
-        python3-dev gcc musl-dev zlib-dev jpeg-dev
-
+        python3-dev gcc musl-dev zlib-dev jpeg-dev && \
 ## Install python dependencies
-RUN python3 -m ensurepip && \
+    python3 -m ensurepip && \
     rm -r /usr/lib/python*/ensurepip && \
-    mkdir -p /usr/src/paperless
-WORKDIR /usr/src/paperless
-COPY requirements.txt /usr/src/paperless/
-RUN pip3 install --no-cache-dir -r requirements.txt
-
-# Copy application
-RUN mkdir -p /usr/src/paperless/src
-RUN mkdir -p /usr/src/paperless/data
-RUN mkdir -p /usr/src/paperless/media
-COPY src/ /usr/src/paperless/src/
-COPY data/ /usr/src/paperless/data/
-COPY media/ /usr/src/paperless/media/
-
-# Set consumption directory
-ENV PAPERLESS_CONSUMPTION_DIR /consume
-RUN mkdir -p $PAPERLESS_CONSUMPTION_DIR
-
+    mkdir -p /usr/src/paperless && \
+    pip3 install --no-cache-dir -r requirements.txt && \
+# Remove build dependencies
+    apk del .build-dependencies && \
+# Create the consumption directory
+    mkdir -p $PAPERLESS_CONSUMPTION_DIR && \
 # Migrate database
-WORKDIR /usr/src/paperless/src
-RUN ./manage.py migrate
-
+    ./src/manage.py migrate && \
 # Create user
-RUN addgroup -g 1000 paperless \
-    && adduser -D -u 1000 -G paperless -h /usr/src/paperless paperless \
-    && chown -Rh paperless:paperless /usr/src/paperless
-
-# Set export directory
-ENV PAPERLESS_EXPORT_DIR /export
-RUN mkdir -p $PAPERLESS_EXPORT_DIR
-
+    addgroup -g 1000 paperless && \
+    adduser -D -u 1000 -G paperless -h /usr/src/paperless paperless && \
+    chown -Rh paperless:paperless /usr/src/paperless && \
+    mkdir -p $PAPERLESS_EXPORT_DIR && \
 # Setup entrypoint
-COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
-RUN chmod 755 /sbin/docker-entrypoint.sh
-
+    chmod 755 /sbin/docker-entrypoint.sh
+WORKDIR /usr/src/paperless/src
 # Mount volumes
 VOLUME ["/usr/src/paperless/data", "/usr/src/paperless/media", "/consume", "/export"]
-
 ENTRYPOINT ["/sbin/docker-entrypoint.sh"]
-
-# Remove build dependencies
-RUN apk del .build-dependencies
 CMD ["--help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,9 @@
 FROM alpine:3.7
 
+LABEL maintainer="The Paperless Project https://github.com/danielquinn/paperless" \
+      contributors="Guy Addadi <addadi@gmail.com>, Pit Kleyersburg <pitkley@googlemail.com>, \
+        Sven Fischer <git-dev@linux4tw.de>"
+
 # Copy application
 COPY requirements.txt /usr/src/paperless/
 COPY src/ /usr/src/paperless/src/
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,10 @@ FROM alpine:latest
 
 # Install dependencies
 RUN apk --no-cache --update add \
-        python3 python3-dev gcc musl-dev gnupg zlib-dev jpeg-dev libmagic \
-        sudo tesseract-ocr imagemagick ghostscript unpaper
+        python3 gnupg libmagic bash \
+        sudo tesseract-ocr imagemagick ghostscript unpaper && \
+    apk --no-cache add --virtual .build-dependencies \
+        python3-dev gcc musl-dev zlib-dev jpeg-dev
 
 ## Install python dependencies
 RUN python3 -m ensurepip && \
@@ -46,4 +48,7 @@ RUN chmod 755 /sbin/docker-entrypoint.sh
 VOLUME ["/usr/src/paperless/data", "/usr/src/paperless/media", "/consume", "/export"]
 
 ENTRYPOINT ["/sbin/docker-entrypoint.sh"]
+
+# Remove build dependencies
+RUN apk del .build-dependencies
 CMD ["--help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,17 @@
-FROM python:3.5
-MAINTAINER Pit Kleyersburg <pitkley@googlemail.com>
+FROM alpine:latest
 
 # Install dependencies
-RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-        sudo \
-        tesseract-ocr tesseract-ocr-eng imagemagick ghostscript unpaper \
-    && rm -rf /var/lib/apt/lists/*
+RUN apk --no-cache --update add \
+        python3 python3-dev gcc musl-dev gnupg zlib-dev jpeg-dev libmagic \
+        sudo tesseract-ocr imagemagick ghostscript unpaper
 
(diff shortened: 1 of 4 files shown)


danielquinn:
I tend to favour the Alpine containers myself, but as @pitkley has been handling the Docker stuff for this project, I've been leaving this sort of thing up to him.  What do you think Pit?

If Pit agrees, I'm happy to merge this, but I wonder if it might be better based on the standard [python:alpine container](https://github.com/docker-library/python/blob/1d59eb2dd813c64891bf554a8ea01754aba25816/3.6/alpine3.4/Dockerfile)?

addadi:
The standard python:alpine container includes libraries that I didn't see that were used in this project.

danielquinn:
Ah, now that I've looked at the Dockerfile I linked, I see what you mean.  Alright, I'll wait on Pit's answer on this one, but it looks ok to me.

pitkley:
Thanks for this PR, and thanks for the mention @danielquinn.

@addadi you say this saves about 400 MB in size -- what is the resulting size? About 480-490 MB?

The reason I ask is that I have [recently started](https://github.com/pitkley/docker-paperless/blob/dfbd8864b21d87205b9dc8352deaae123ef3a28b/Dockerfile) to convert the Dockerfile to use the Alpine-version of the python base image, and I achieve a size of 358 MB in comparison to the 891 MB of the current image on Docker Hub (as given by `docker image ls`).

I would personally prefer to be based on the upstream managed python base image, rather than managing the python installation ourselves. But if we can reduce the size further by not relying on the packages used upstream, it might be worthwhile.

addadi:
after my last commit  the image size is at 416 MB. Your image is therefore more compact, but the it seems like the dockerfile is more complex, despite being based on the python base image. I'll try to tweak the command to see if the size can be reduced even further, but either version of the dockerfiles, it's already better than the current dockerfile based on debian

pitkley:
One of the reasons your Dockerfile is bigger (and the one I posted seems more complex), is that I have reduced the image down to a single `RUN` statement.

You are [adding](https://github.com/addadi/paperless/blob/7d81de4edf7bf540457ea4f3e984c5b3a9b9151b/Dockerfile#L4-L8) the dependencies in one `RUN` statement, and [removing](https://github.com/addadi/paperless/blob/7d81de4edf7bf540457ea4f3e984c5b3a9b9151b/Dockerfile#L53) it in another. This will not cause the size of the image to go down, but rather to go up ever so slightly, since Docker images work with layers. Every file you remove in a later layer will still be there in the layer that came earlier -- and it will be downloaded by clients and thus counts against the image size.

Strubbl:
I want to point out, that i do not like using "alpine:latest". I would like to use "alpine:3.7" instead of latest. This way we do not get a broken image someday when a new alpine major is released. 

When a new alpine major version comes out, we can always update the Dockerfile.

addadi:
The updated version is based on alpine:3.7 which includes just one layer of RUN commands. Also I've removed some redundant mkdir commands whic makes the docker image size just 326MB and the dockerfile easier to maintain and to read. Thanks all for the valuable feedbacks

danielquinn:
Looks good to me.  As more-senior people on this project, @pitkley, @Strubbl what do you think?

pitkley:
I'll try to review ASAP, but it might take a day or two.

Dockerfile
```diff
@@ -1,50 +1,41 @@
-FROM python:3.5
-MAINTAINER Pit Kleyersburg <pitkley@googlemail.com>
```
Strubbl: Shouldn't we leave in the maintainer label? (or upgrade the line according to https://docs.docker.com/engine/reference/builder/#maintainer-deprecated )
I like having a point of contact in the dockerfile itself.

pitkley: I personally don't mind the maintainer gone, but if we keep it, then we should definitely use the `LABEL` instruction.

If we keep it, we should also maybe update it to reference this repository rather than a single person, since this never has been only my effort, but rather that of multiple great people.

*If* you add the `LABEL`, please add it as the last instruction in the `Dockerfile`, such that a update of the label will not require the whole image to be rebuilt.

Dockerfile
```diff
@@ -1,50 +1,41 @@
...
 # Copy application
-RUN mkdir -p /usr/src/paperless/src
-RUN mkdir -p /usr/src/paperless/data
-RUN mkdir -p /usr/src/paperless/media
 COPY src/ /usr/src/paperless/src/
 COPY data/ /usr/src/paperless/data/
 COPY media/ /usr/src/paperless/media/
-
+# Set export directory
+ENV PAPERLESS_EXPORT_DIR /export
```
pitkley: Since we are trying to reduce the size of the image, we should also strive to reduce the amount of layers.

I would suggest to combine this and the next `ENV` instruction down to one, see [here](https://github.com/pitkley/docker-paperless/blob/dfbd8864b21d87205b9dc8352deaae123ef3a28b/Dockerfile#L3-L5) for how it works.

Dockerfile
```diff
@@ -1,50 +1,41 @@
...
-RUN mkdir -p $PAPERLESS_CONSUMPTION_DIR
-
+COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
+# Install dependencies
+RUN apk --no-cache --update add \
+        python3 gnupg libmagic bash \
+        sudo tesseract-ocr imagemagick ghostscript unpaper && \
+    apk --no-cache add --virtual .build-dependencies \
+        python3-dev gcc musl-dev zlib-dev jpeg-dev && \
+## Install python dependencies
```
pitkley: Minor nit: Remove one of the `#`.

Dockerfile
```diff
@@ -1,50 +1,41 @@
...
+    addgroup -g 1000 paperless && \
+    adduser -D -u 1000 -G paperless -h /usr/src/paperless paperless && \
+    chown -Rh paperless:paperless /usr/src/paperless && \
+    mkdir -p $PAPERLESS_EXPORT_DIR && \
 # Setup entrypoint
-COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
-RUN chmod 755 /sbin/docker-entrypoint.sh
-
+    chmod 755 /sbin/docker-entrypoint.sh
+WORKDIR /usr/src/paperless/src
```
pitkley: General comment: please reintroduce blank lines into the `Dockerfile`. Having everything as crammed as it is right now makes it a lot harder -- at least for me -- to identify the different logical groupings that were in the file before, e.g. having the `COPY`s visually separated from the `ENV` and so on.

Dockerfile
```diff
@@ -1,50 +1,41 @@
...
 # Mount volumes
 VOLUME ["/usr/src/paperless/data", "/usr/src/paperless/media", "/consume", "/export"]
```
pitkley: General comment: please reintroduce blank lines into the `Dockerfile`. Having everything as crammed as it is right now makes it a lot harder -- at least for me -- to identify the different logical groupings that were in the file before, e.g. having the `COPY`s visually separated from the `ENV` and so on.

Dockerfile
```diff
@@ -1,50 +1,41 @@
...
-RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-        sudo \
-        tesseract-ocr tesseract-ocr-eng imagemagick ghostscript unpaper \
-    && rm -rf /var/lib/apt/lists/*
-
-# Install python dependencies
-RUN mkdir -p /usr/src/paperless
+FROM alpine:3.7
 WORKDIR /usr/src/paperless
```
pitkley: In light of reducing the image layers, please remove this `WORKDIR` instruction. This requires a change later on in the `RUN` instruction, I have added a comment on what I think is required.

Dockerfile
```diff
@@ -1,50 +1,41 @@
...
+RUN apk --no-cache --update add \
+        python3 gnupg libmagic bash \
+        sudo tesseract-ocr imagemagick ghostscript unpaper && \
+    apk --no-cache add --virtual .build-dependencies \
+        python3-dev gcc musl-dev zlib-dev jpeg-dev && \
+## Install python dependencies
+    python3 -m ensurepip && \
+    rm -r /usr/lib/python*/ensurepip && \
+    mkdir -p /usr/src/paperless && \
+    pip3 install --no-cache-dir -r requirements.txt && \
```
pitkley: In reference to the other comment regarding the removal of the one `WORKDIR` instruction, it probably is enough to add a `cd /usr/src/paperless` between the `mkdir` and `pip3` commands.

The latter `WORKDIR` instruction ensures that we execute from the right directory.

addadi:
@pitkley thanks for your review. I've changed few things according to your feedback and I'm going to push another version. Keep in mind the following:
1. We can reduce indeed the number of layers but it won't change the image size. The reason is that these ENV WORKDIR etc layers do not contain commands that consume space. We will still be in 326 MB. I still am going to remove and merge them wherever possible as I'd prefer to have a minimal and simple Dockerfile
2. The blank lines were removed because they generate warning of "Empty continuation lines will become errors in a future release." . I can reintroduce them in the lines that are not the part of the RUN command.
3. As for the maintainer - Just a clarification I originally removed it as this started as my personal improvement for the image and I didn't need this as I prefer to keep the file as minimal as possible. You can decide to reintroduce it of course.
I have verified of course that paperless is working after each build. The containers are up and running on my pc and I'm able to work with the web server but I haven't verified each and feature of paperless. I'd also like to introduce in the future a helper script that take the OCR languages from the docker-compose.env and install the appropriate tesseract package, but I'll do it once I have time for it.
Thanks again for the comments. I'm going to push the updated file soon

addadi:
I've done some additional minor changes such as adapting the docker-compose.yml file to work with the new docker image and changing the entrypoint script to install additional OCR languages. I've been working with this for few days and it seems to me that all functionality is fine + of course the image file size is significantly smaller. Anybody else tested it? can it be merged already?
Thanks

danielquinn:
Hello all, so what's the consensus on this one?

addadi:
I've checked the upgrade path if coming from an earlier version (debian-based) of the docker-compose images to the newer one based on alpine. it's pretty straight forward except a minor change in the docker-compose.yml (move from pull an image from docker jhub to build an image based on alpine). The needed change is documented in the setup.rst.

danielquinn:
@pitkley I just tried to create a `Paperless` organisation, but the form just outlines the box in red, so I assume the namespace is taken?  Which is weird, since [this is a 404](https://hub.docker.com/_/paperless/).  Did I miss something?

pitkley:
@danielquinn [Apparently there is a user called `paperless`](https://hub.docker.com/r/paperless/), which means that the namespace is taken. This would probably only leave option 1 of having you host the image under `danielquinn/paperless` -- I suppose that is as official as it will get.

danielquinn:
And they're not even using it... BOO!

Alright well I started experimenting with Docker Hub a long time ago and have had an automated build at [danielquinn/paperless](https://hub.docker.com/r/danielquinn/paperless/) for a while now, though I don't know how well it works ;-)

Alright given the above discussion, it looks like we've got a consensus here.  We're moving to Alpine!  Merging now :-)

@pitkley I appreciate you hosting the "semi-official" docker instance all this time, but I think you're right that it deserves an "official" home.  Would setting up a `paperless-official` organisation be a better choice, or perhaps another name, should one of you have a better idea?

pitkley:
> though I don't know how well it works ;-)

Given that the [last build is 9 months old](https://hub.docker.com/r/danielquinn/paperless/builds/), probably not at all. :smile:

The build settings for the Docker Hub repository should look similar to this:

s2018-01-30_18-20-47

> it deserves an "official" home

Since both the GitHub and Docker Hub repositories are named identical, I don't think there is a huge issue here. But I concur, if somebody can think of a good name, switching to that might be a good choice. This should happen fairly quickly though, before a bunch of users use `danielquinn/paperless` and we effectively split the users into three "official" images:

* `pitkley/paperless`
* `danielquinn/paperless`
* `?/paperless`

danielquinn:
This is the thing I don't get about Docker Hub.  They have features for automatic building, but when I turn them on, *they don't actually run*.  This is what my screen looks like (and has looked like for months):

s2018-01-30_18-20-47 

...and yet, nothing is built unless I click the `Trigger` button there.  I've now *deleted the hub page for danielquinn/paperless* and re-created it, specifically as an automated build (in case I somehow didn't choose that last time) and yet, even after pushing an update to the repo, *nothing is being built*.  😡 

I tried a few other options for starting an organisation: `paperless_official`, and `paperless-official`, both failed, so I think we're stuck with `danielquinn/paperless`.  Anyone care to help me with what I'm missing?

danielquinn:
Ah.  [This appears to be the problem](https://github.com/docker/hub-feedback/issues/813#issuecomment-336610700).  I'll see what I can do to fix this.

danielquinn:
Ok I've figured out how to get Docker Hub to auto-update, and reworked the README file to be Docker-friendly (they really don't make it easy), so things appear to be [stable now](https://hub.docker.com/r/danielquinn/paperless/).  @pitkley: any objections to me changing the documentation to point to `danielquinn/paperless` instead of `pitkley/paperless`?

danielquinn:
Strike that, I just checked, and it turns out that we don't reference Docker Hub in the documentation, opting instead to tell them to just use `build: ./`.  So there's nothing to change :-D

PR #277, opened 2017-12-23
Multi word match

ishirav:
Hi Daniel,

Thank you for the Paperless project, it's exactly what I was looking for and it's extremely well-built.

Here's a pull request for your consideration: support for multi-word search terms in match text.

Use case: suppose I want to add Bob Marley as a correspondent, but his name may appear either as "Bob Marley" or "Marley, Bob". So with this feature, I can use the MATCH_ANY mode and set the match text to `"Bob Marley" "Marley, Bob"`.

### DIFF
--- a/docs/guesswork.rst
+++ b/docs/guesswork.rst
@@ -80,6 +80,12 @@ text and matching algorithm.  From the help info there:
     uses a regex to match the PDF.  If you don't know what a regex is, you
     probably don't want this option.
 
+When using the "any" or "all" matching algorithms, you can search for terms that
+consist of multiple words by enclosing them in double quotes. For example, defining
+a match text of ``"Bank of America" BofA`` using the "any" algorithm, will match
+documents that contain either "Bank of America" or "BofA", but will not match
+documents containing "Bank of South America".
+
 Then just save your tag/correspondent and run another document through the
 consumer.  Once complete, you should see the newly-created document,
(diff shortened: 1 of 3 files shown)


danielquinn:
I think this may be the cleanest pull request this project has ever seen.  Nice work, and thanks for your contribution!

danielquinn:
@ishirav I just ran the tests with `--pythonwarnings=all` and I got this DeprecationWarning from the code you added in this PR:

```
documents/tests/test_matchables.py::TestMatching::test_match_all
  /home/daniel/Projects/Paperless/src/documents/models.py:136: DeprecationWarning: bad escape \s
    return [normspace(r"\s+", (t[0] or t[1]).strip()) for t in findterms(self.match)]
```

The complaint appears to be coming from a list comprehension you added containing a `\s+` [here](https://github.com/danielquinn/paperless/blob/master/src/documents/models.py#L135), but I don't really understand what you're trying to do there.  Can you help me out?

ishirav:
The idea behind this line is to build a regular expression, such as "Daniel\s+Quinn" so that any whitespace characters will be acceptable between the words.
One way to get rid of the warning:
```
return [normspace(" ", (t[0] or t[1]).strip()).replace(" ", r"\s+") for t in findterms(self.match)]
```

danielquinn:
Excellent.  Thank you :-)

PR #285, opened 2018-01-19
small typo in exporter thumbnail filename

pzl:
The typo doesn't affect functionality, since whatever filename is used gets written to `manifest.json`, and that is what is used to import. So this is purely cosmetic.


danielquinn:
That typo has been there *for ages*!  Thanks!

PR #291, opened 2018-01-28
Add support for a heuristic that extracts the document date from its text

BastianPoe:
If the date of a document is not given in its filename, this patch will try to heuristically extract the date of the document by looking at the text of it.


danielquinn:
This is excellent, thank you.  I do have one question though: you've elected to use `dateparser`, but I'm curious if this can be done without injecting a new dependency.  Is it possible that `dateutil` could do what you're doing here?  I admit I'm unfamiliar with dateparser.

src/paperless_tesseract/parsers.py
```diff
@@ -191,6 +193,29 @@ def _assemble_ocr_sections(self, imgs, middle, text):
...
+        # - MONTH ZZZZ
+        m = re.search(
+            r'\b([0-9]{1,2})[\.\/-]([0-9]{1,2})[\.\/-]([0-9]{4}|[0-9]{2})\b|' +
+            r'\b([0-9]{1,2}\. [^ ]{3,9} ([0-9]{4}|[0-9]{2}))\b|' +
+            r'\b([^ ]{3,9} [0-9]{4})\b', text)
+
+        if m is None:
+            return None
+
+        return dateparser.parse(m.group(0),
```
danielquinn: Ok so I did some digging around on options here re dateutil vs dateparser and I *think* we can do without it if we're a little more creative.  Given that we've already got the date string pre-parsed out for us, we can use dateutil's `parse()` function to do most of the heavy lifting:

```python
from dateutil import parse
return parse(m.group(0), dayfirst=True, yearfirst=True)
```

The `dayfirst=` and `yearfist=` values can be determined by the `DATE_ORDER` setting you setup above.  Something like:

```python
kwargs = {"dayfirst": settings.DATE_ORDER == "MDY", "yearfirst": False}
return parse(m.group(0), **kwargs)
```

Finally, the returned value needs to be made timezone aware, which dateutil won't do for us, but which we can do for ourselves easily enough:

```python
date = parse(m.group(0), **kwargs)
if date.tzinfo is None:
    date = pytz.timezone("UTC").localize(date)
return date
```

What do you think?

danielquinn:
So now this PR has a conflict because of the merging of your *other* PR :-)

I had a little time today so I've left one comment about how I think we could do this without adding the dateparser depdendency, so would you mind having a look when you've got a minute to see if this is feasible?  Outside of that, the only thing this is missing is some tests.  This is the sort of thing that could benefit from the kind of test that takes a bunch of dummy text and pulls out the right date, handles timezone stuff etc.  If you've no time for that though, I can write it.

This looks good, and I'm happy to merge it once we've got this bit sorted.

BastianPoe:
Ok, did the changes as proposed. The only downside is, that we now need to have our own file for localizing the date detection as dateutil only knows about English. I have added German (which is my use case) but other languages will have to be added later if required. The settings default to English, as you would expect.

danielquinn:
I see now why you started with dateparser.  Handling localisation is a pain and offloading that to a library is usually a good idea.  My problem was that I assumed that your regex was only looking for *numbers*, and so localisation didn't seem like a problem we needed to address.  I see now that your code is actually smart enough to be capturing names as well, so I'm so sorry I made a big stink about adding a new module.

You were right the first time around.  Let's add `datetparser` to the requirements and go with your original plan (if you don't mind reverting).  Taking on a new dependency makes a lot more sense than assuming on the responsibility of handling all of the localisations ourselves.

As a side note, the Travis failure appears to be the result of an update to Pytest that was just released that conflicts with one of the other test dependencies (pytest-sugar).  I've since fixed this, so the tests *should* pass now.

maur:
Great addition for scanning old documents, like when moving archive to e-archive on paperless. Amazing work, thank you guys!

BastianPoe:
Ok, I have now reverted the latest changes and added some test cases including test documents. Furthermore, I have done some performance optimizations.

How do you feel about the pull request?

danielquinn:
@BastianPoe this is outstanding. Well above-and-beyond what's typically expected from a pull-request.  Thank you for your excellent work.  Merging now!

PR #294, opened 2018-02-01
Fixing error sentinel for pdftotext when the PDF has no text (scanned…

matthewmoto:
… images). It was causing a crash previously.


danielquinn:
Good catch :-)

Issue #299, opened 2018-02-04
Docker install from scratch doesn't run migrate


TeraHz:
Just installed paperless for the first time from scratch in docker, and the createsuperuser step was failing due to missing tables. There was a warning about migrations that needed to be done. This was also causing the consumer container to fail to start. I ran the migrations and everything started up nicely. 

Off to feed it some PDFs now to see how it does!

danielquinn:
This is odd because [there's a line in the Dockerfile to run migrations](https://github.com/danielquinn/paperless/blob/master/Dockerfile#L34).  I'm not much of a Docker pro, but I just did the following and it worked for me:

1. Cloned this repo
2. Added two files:
    `docker-compose.env`:
    PAPERLESS_PASSPHRASE=CHANGE_ME
    PAPERLESS_OCR_THREADS=16
    PAPERLESS_OCR_LANGUAGES=eng nld
    `docker-compose.yml`:
        
        version: '2'
        
        services:
            webserver:
                build: ./
                ports:
                    - "8000:8000"
                volumes:
                    - data:/usr/src/paperless/data
                    - media:/usr/src/paperless/media
                env_file: docker-compose.env
                environment:
                    - PAPERLESS_OCR_LANGUAGES=
                command: ["runserver", "--insecure", "0.0.0.0:8000"]
        
            consumer:
                build: ./
                volumes:
                    - data:/usr/src/paperless/data
                    - media:/usr/src/paperless/media
                    - /tmp/paperless/consume:/consume
                    - /tmp/paperless/export:/export
                env_file: docker-compose.env
                command: ["document_consumer"]
        
        volumes:
            data:
            media:
3. Ran `docker-compose up`

This resulted in a lot of output, but importantly it included this section:
```
Operations to perform:
  Apply all migrations: admin, auth, contenttypes, documents, reminders, sessions
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying admin.0001_initial... OK
  Applying admin.0002_logentry_remove_auto_add... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
  Applying auth.0004_alter_user_username_opts... OK
  Applying auth.0005_alter_user_last_login_null... OK
  Applying auth.0006_require_contenttypes_0002... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
[...]

PR #300, opened 2018-02-05
Handle Django migrations at startup

TeraHz:
Here is one solution to https://github.com/danielquinn/paperless/issues/299.
Basically, I've established a dependency between the two containers and moved the migration to the entrypoint. Startup takes a few more seconds now, while the health check is happening. 

I've also updated the docs and added an example restart policy to the compose example file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,8 +30,6 @@ RUN apk --no-cache --update add \
     apk del .build-dependencies && \
 # Create the consumption directory
     mkdir -p $PAPERLESS_CONSUMPTION_DIR && \
-# Migrate database
-    ./src/manage.py migrate && \
 # Create user
     addgroup -g 1000 paperless && \
     adduser -D -u 1000 -G paperless -h /usr/src/paperless paperless && \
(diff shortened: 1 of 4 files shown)


danielquinn:
@TeraHz I like this, but I've been proven to have some questionable judgement when it comes to Docker stuff in the past, so I try to run these sorts of changes by @pitkley and @addadi for a second opinion.  I understand they're a little swamped right now (I know the feeling), so this make take a while to merge.  It'll get there eventually though :-)

TeraHz:
No worries. I'm running with that change and the only "downside" to this code is that it runs migrations every time you start the container (even if you run retagging, for example). So it adds a couple of seconds to the runtime. But ensures things are in order. 

I'm also itching to convert it to a proper deployment with nginx/gunicorn etc, but will probably wait a bit on that :)

TeraHz:
Oops, my bad - retabbed with expandtab. Weird, the 8 space indents were tabs, the 12 space indents were spaces... no idea how that happened. 

As far as testing it - yes. I'm running this way now and I've rebuilt and restarted multiple times since then.

danielquinn:
Thanks to both of you for working this out!  Merging now.

Issue #301, opened 2018-02-07
Regex for OCR'd dates matching on invalid date format for dateparser


isaacsando:
I have a PDF that contains this string:
December 31, 2017

When the document_consumer code runs and processes this PDF it throws the following exception (trucated):
```python
  File "/home/isaac/devel/python_projects/paperless/src/paperless_tesseract/parsers.py", line 224, in get_date
    'RETURN_AS_TIMEZONE_AWARE': True})
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/conf.py", line 84, in wrapper
    return f(*args, **kwargs)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/__init__.py", line 40, in parse
    data = parser.get_date_data(date_string, date_formats)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/date.py", line 371, in get_date_data
    language, date_string, date_formats, settings=self._settings)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/date.py", line 168, in parse
    return instance._parse()
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/date.py", line 178, in _parse
    date_obj = parser()
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/date.py", line 199, in _try_parser
    self._get_translated_date(), settings=self._settings)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/conf.py", line 84, in wrapper
    return f(*args, **kwargs)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/date_parser.py", line 26, in parse
    date_obj, period = parse(date_string, settings=settings)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/parser.py", line 84, in parse
    raise exceptions.pop(-1)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/parser.py", line 78, in parse
    res = parser(datestring, settings)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/parser.py", line 454, in parse
    po = cls(tokens.tokenize(), settings)
  File "/home/isaac/.venv/paperless/lib/python3.6/site-packages/dateparser/parser.py", line 288, in __init__
    datetime(**params)
[...]

PR #302, opened 2018-02-08
Extends the regex to find dates in documents as reported by @isaacsando

BastianPoe:
This extends the regex to find dates in documents as reported in #301 

@isaacsando, please try out and report back


ddddavidmartin:
One thought on the date matching in general:
It would be great if the detected date was printed in the log output. This would make it easier to sanity-check that the correct date was applied, and to find the document again if not. For example I had it happen that my birth date was the first in the document and it was stored under that.

Feel free to ignore if this is out of scope of this pull request.

BastianPoe:
@ddddavidmartin please have a look if the change matches want you wanted.

ddddavidmartin:
@BastianPoe, I have not had a chance yet to actually run this branch, but looking at https://github.com/danielquinn/paperless/pull/302/commits/c74bb84c8321a03feb5e2a128bb8764d5795f783 this is exactly what I was asking for. Thanks!

danielquinn:
I'm inclined to merge this.  @BastianPoe & @isaacsando you cool with that?

danielquinn:
I just saw that @isaacsando gave this the thumbs up in #301 so I'm merging it now :-)

Issue #306, opened 2018-02-17
Docker: group 'paperless' in use


ChromoX:
When attempting to start the ```danielquinn/paperless``` I get the following error message:
```
Mapping UID and GID for paperless:paperless to 1008:1008
addgroup: group 'paperless' in use
```

Any help would be greatly appreciated.

danielquinn:
Yikes.  It looks like you've run up on a bug that only presents itself when the user running the container has a UID other than `1000`.  In your case, you've got `1008` and that's triggering [this line block of code]() that attempts to create the already-existing `paperless` group.

I'm curious as to the right methodology to follow here, so I'm going to reach out to @pitkley and @addadi on this one.  The problem is in two spots:

1. We create the `paperless` group in the Dockerfile, so creating it here doesn't make sense any more.
2. We create the group with a hard-coded id of `1000`, which wouldn't work very well for people like @ChromoX here.

The question then is whether there's a way to dynamically set the GID in the Dockerfile (I'm not sure how that'd work as it's built automatically), if we should move the user/group creation into the entrypoint, or if the entrypoint should just have some code to change the UID/GIDs -- though I'm unsure of how to do this in Alpine.

Any thoughts guys?

ChromoX:
@danielquinn Thanks for the information. I can simply set my UID and GID to 1000 for now.

LinuxServer.io containers allow for setting the UID and GID from environment variables. Later tonight I can take a shot at integrating how they do it with paperless and issue a pull request.

PR #311, opened 2018-02-22
Fix UID & GID Mapping on Docker container startup

ChromoX:
This fixed #306 for me.

Fixed a typo as well as there were two ```USERMAP_ORIG_UID```s.
Removed some hanging white space as well.

PR #312, opened 2018-02-24
Fix modifying the GID under alpine

pitkley:
This fixes issue #306.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ ENV PAPERLESS_EXPORT_DIR=/export \
 
 # Install dependencies
 RUN apk --no-cache --update add \
-        python3 gnupg libmagic bash \
+        python3 gnupg libmagic bash shadow \
         sudo poppler tesseract-ocr imagemagick ghostscript unpaper && \
     apk --no-cache add --virtual .build-dependencies \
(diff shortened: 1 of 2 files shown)


pitkley:
I have only just seen #311. I've updated my PR to include the typos @ChromoX fixed -- thanks for that!

Pro @ChromoX solution: no additional package required.
Pro from my solution: no need to delete and recreate both the user and the group.

I don't care which one gets merged, both should be fine! :slightly_smiling_face:

ChromoX:
I prefer your commit as it's obvious to what the intention of the code is.

pitkley:
Okay, thanks for the consensus. I think there is still an issue with mapping the UID, I'll push a fix shortly. I'll ping you @danielquinn once the PR is ready!

pitkley:
I have updated the PR to use `usermod` since it is available now that we have `shadow` installed (thanks @ChromoX!).

I have also renamed the computed `USERMAP_UID` and `USERMAP_GID` to `USERMAP_NEW_UID` and `USERMAP_NEW_GID` respectively. I hope this makes the logic more clear.

This PR is ready from my side @danielquinn!

danielquinn:
Hooray!  Thanks to both of you for your hard work (and collaboration) on this!

Issue #305, opened 2018-02-15
Expose cli argument for `document_consumer`


ovv:
Hello,

Would exposing arguments for the `document_consumer` script be ok with you ?

My primary goal is to allow document importing from two directories, this could also be done via the `paperless.conf` file but allowing argument would permit deeper configuration in the future (languages, default tags, default correspondent, ...).

danielquinn:
I don't see why not.  So long as things continue to run properly *without* command-line arguments (defaulting to `paperless.conf` values) then that's cool with me.  You offering to write this?

ovv:
Yeah I can give it a shot.

danielquinn:
Cool beans!  Consider it yours :-)

PR #313, opened 2018-02-24
Configuration cli argument for document_consumer

ovv:
See #305 

Putting this out to get early feedback

### DIFF
--- a/docs/utilities.rst
+++ b/docs/utilities.rst
@@ -80,6 +80,13 @@ you'll need to have it start in the background -- something you'll need to
 figure out for your own system.  To get you started though, there are Systemd
 service files in the ``scripts`` directory.
 
+Some command line argument are available to customize the behavior of the
+consumer. By default it will use ``/etc/paperless.conf`` values. Display the
+help with:
+
+.. code-block:: shell-session
+
+    $ /path/to/paperless/src/manage.py document_consumer --help
 
 .. _utilities-exporter:
 
--- a/docs/utilities.rst
+++ b/docs/utilities.rst
@@ -80,7 +80,7 @@ you'll need to have it start in the background -- something you'll need to
 figure out for your own system.  To get you started though, there are Systemd
 service files in the ``scripts`` directory.
 
-Some command line argument are available to customize the behavior of the
+Some command line arguments are available to customize the behavior of the
 consumer. By default it will use ``/etc/paperless.conf`` values. Display the
 help with:
 
--- a/src/documents/management/commands/document_consumer.py
+++ b/src/documents/management/commands/document_consumer.py
@@ -30,17 +30,25 @@ class Command(BaseCommand):
         BaseCommand.__init__(self, *args, **kwargs)
 
     def add_arguments(self, parser):
-        parser.add_argument("directory", default=settings.CONSUMPTION_DIR, nargs='?')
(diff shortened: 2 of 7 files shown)


src/documents/management/commands/document_consumer.py
```diff
@@ -32,13 +29,22 @@ def __init__(self, *args, **kwargs):
...
+    def add_arguments(self, parser):
+        parser.add_argument("directory", default=settings.CONSUMPTION_DIR, nargs='?')
+        parser.add_argument("--loop-time", default=settings.CONSUMER_LOOP_TIME, type=int)
+        parser.add_argument("--mail-delta", default=10, type=int)
+        parser.add_argument("--oneshot", action='store_true')
+
     def handle(self, *args, **options):
 
         self.verbosity = options["verbosity"]
+        directory = options['directory']
```
danielquinn: Please stick to using `"` for strings unless those strings contain a `"` themselves

danielquinn:
Hey @ovv, just a heads-up that I merged a change this morning that moves `pytest.ini` to `setup.cfg` (along with a slew of other things like Coveralls support).  As a result, your branch now has a conflict, but if you stash your changes & rebase off master, you'll only need to re-remove that one line from `setkup.cfg`.

ovv:
This should be ok for review @danielquinn 

I am a bit confused why you decided to remove tox. It's quite an useful tool for testing on multiple python version without waiting on Travis. It should handle out of the box settings in `setup.cfg`.

docs/utilities.rst
```diff
@@ -80,6 +80,13 @@ you'll need to have it start in the background -- something you'll need to
 figure out for your own system.  To get you started though, there are Systemd
 service files in the ``scripts`` directory.
 
+Some command line argument are available to customize the behavior of the
```
danielquinn: The only thing I could find that needs attention is this typo here :-)  It should probably be `Some comand line arguments`.

danielquinn:
This is some really good code, and *pretty* code too!  Thank you so much for this contribution!  I have just that one typo nitpick to be fixed and then I can merge it.

As for tox, I think you're right on that.  I was a bit overzealous in removing it when I was trying to push more into Travis.  I'll put it back when I have some time unless you want to do that in this PR or a separate one.

ovv:
Thanks for the kind words. It should be good now.

danielquinn:
So I downloaded your code and ran it without issue, save for one problem: the `--oneshot` flag ran the `consumer.run()` alright, but the consumer loop has to run *at least twice* for anything to happen since the first run just takes file statistics to ensure we don't consume a file that's currently being written to.

I made that one modificaiton and then merged all of your code into master.  Thank you so much for this excellent body of work!  It's annoying how Github now says that this branch has conflicts, but I guess that's what happens when I tweak something before a merge.  Oh well.

I'll be doing a release this evening with your modifications.  Thanks again!

danielquinn:
On second thought, I think I'd like to make sure #317 isn't a problem before I release.  If you're just running a checkout of `master` though, a `git pull` will have you running the latest.  Otherwise, I'm going to wait to see what happens with that thread before I release 1.4.0.

ovv:
Good catch ! Thanks for the fix & merge. A new release is not urgent for me, I'm running on master at the moment.

PR #314, opened 2018-02-25
WIP: Add Coveralls

danielquinn:
Refactor the various .ini files into two files: `.travis.yml` and `setup.cfg`.  This PR also starts us down the road toward using a Pipfile instead of requirements.txt


danielquinn:
@BastianPoe do you mind having a look at what I've done here with Travis & Coveralls?  I've tried to do away with tox, but I want to make sure I'm doing this right.

BastianPoe:
Looks good to me so far - and the coverage is not as bad as I thought based on what you said. Also, I have already learned the first things about my tests and how to improve them.

Do you feel ok with failing the commit whenever the coverage is decreased? I guess this will have a motivating effect on all contributors. ;)

danielquinn:
You know, I'm not sure how I feel about having it fail for reduced coverage, but so long as I can still say "fuck it, merge anyway", I don't see why we can't start from that position and see how it affects users.  The thing is, I'd love to see more test coverage, but I find the output of coverage cryptic and often can't figure out what it thinks I need to test.

Let's start with it flagging reduced coverage and see how the other devs react.  Frankly, I don't understand the output of Coveralls, but as it's a popular tool (and you find it useful) I'm happy to add it if it means we get more tests.

PR #316, opened 2018-02-27
Add back tox

ovv:
It seems to be running fine with the exact same file as before

### DIFF
--- a/src/tox.ini
+++ b/src/tox.ini
@@ -5,7 +5,7 @@
 
 [tox]
 skipsdist = True
-envlist = py34, py35, py36, pycodestyle
+envlist = py34, py35, py36, pycodestyle, doc
 
 [testenv]
 commands = pytest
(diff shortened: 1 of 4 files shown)


ovv:
It would probably be good to add a test for the documentation build in travis & tox but I'm not sure how it's built at the moment.

src/tox.ini
```diff
@@ -0,0 +1,22 @@
+# Tox (http://tox.testrun.org/) is a tool for running tests
```
danielquinn: Is it possible to roll this file's content into `setup.cfg`?  I'd just like to limit the number `.ini` files littered about if I can help it.

ovv: afaik tox requires a `tox.ini` but it uses the `setup.cfg` and others files to get tools configuration. Like for `pycodestyle`.

danielquinn:
Ah I see now.  There appears to be [quite the controversy](https://github.com/tox-dev/tox/issues/297) about the whole setup.cfg thing so I'm just going to keep my nose out of it.

Thanks so much for the PR!

src/tox.ini
illume: tox doesn't require a tox.ini (now), setup.cfg can be used with a [tox:tox] section. Leaving this here because it's highly ranked on a search engine. :)

PR #320, opened 2018-03-04
Log detected document date with isoformat

ovv:
Logging the document date with the locale specific format can be confusing if the ocr language is not the same as the locale.


danielquinn:
We should probably be using `%s` here rather than concatenating like this, but `.isoformat()` is definitely better regardless.  Thanks :-)

PR #322, opened 2018-03-04
Added support for paperless.conf in /usr/local/etc

Arendtsen:
As a FreeBSD user I prefer to have the config file in /usr/local/etc


danielquinn:
Sure why not :-)

We may want to rethink the idea of a hard-coded path at all, but I see no harm in expanding the location like this.

danielquinn:
Thanks for the PR!

Issue #319, opened 2018-03-04
New imported documents list


ovv:
Following #302 the date matching is working way better for my docs thanks. 

I usually check that the scanned document is of good quality before recycling/archiving the originals, now that the dates are matched it became harder to find them. I think providing a list of recent imports and/or a way to order documents by import date would be useful to check the quality of recent documents.

ddddavidmartin:
Agreed. I think the list of Recent Actions on the start page may be a good spot to mention newly imported documents. 

I was actually considering adding an option to disable the automatic date matching for this reason. It seemed easier to implement. I tend to update the title and tags by hand for new documents anyway so changing the date does not add much effort. Or, less effort than trying to get to the document as it is.

ovv:
The `Recent Actions` seems like a good fit. As far as I know it's a standard part of the django admin so I'm not sure if we can easily add new actions from the `document_consumer`

danielquinn:
It looks like the `Recent Actions` section is handled as part of [Django's `LogEntry` model](https://docs.djangoproject.com/en/2.0/ref/contrib/admin/#logentry-objects).  Adding some code to the consumer to create a LogEntry object might be all that's necessary if one of you wanna try it.

ddddavidmartin:
Thanks for the pointer @danielquinn! I'm pretty busy at the moment and I won't be able to contribute more than a bit of armchair commenting, so feel free to give it a go @ovv. I'm happy to try out any proposed changes though.

ovv:
Not really sure when I can take a look at it but I'll drop a comment when / if I start. In the meantime if anyone want to give it a shit, that's ok with me.

CkuT:
I quickly gave a shot for this :
https://github.com/danielquinn/paperless/commit/45e18d709406e9631c5406d5c718c5ef43de1351

It is just a new signal handler on `document_consumption_finished` that creates a LogEntry object.

https://i.imgur.com/dyTMqZo.png

I'm open to improvements about some little things :
- LogEntry wants a `User` for creating the object. I currently take the first user of the database, but maybe we could create a specific "consumer" user ?
- I use the `document.created` as the `LogEntry.action_time`, but I think `document.modified` would be better. What do you think ?

ddddavidmartin:
Hi @CkuT, nice! Would you be able to open a pull request with your changes? Then they are easy to try out just by checking out the pull request branch. Thanks!

ovv:
Looks good.
[...]

PR #326, opened 2018-03-12
New imported documents list

CkuT:
See #319

### DIFF
--- a/src/documents/apps.py
+++ b/src/documents/apps.py
@@ -15,13 +15,15 @@ class DocumentsConfig(AppConfig):
             set_tags,
             run_pre_consume_script,
             run_post_consume_script,
-            cleanup_document_deletion
+            cleanup_document_deletion,
+            set_log_entry
         )
 
         document_consumption_started.connect(run_pre_consume_script)
 
         document_consumption_finished.connect(set_tags)
         document_consumption_finished.connect(set_correspondent)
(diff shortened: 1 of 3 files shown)


CkuT:
This is a WIP : some tests are not written and other may not pass

CkuT:
Oddly, some tests do not pass if the variable `PAPERLESS_SHARED_SECRET` is not defined.

I think it comes from this part of code : https://github.com/danielquinn/paperless/blob/master/src/documents/mail.py#L46-L49

danielquinn:
This all looks good to me.  Are you cool with me merging it, or did you want to discuss things more on #319?

danielquinn:
Also, regarding the test failure, I re-ran the tests locally and couldn't reproduce the error you mentioned here.  Regardless, I removed the line you pointed too as it was well past time to do so anyway.  Out of curiosity, how are you running the tests exactly?  This is how I do it:

```bash
cd src/
pytest
```

...either that or using tox to be thorough:

```bash
cd src/
tox
```

CkuT:
You can merge it. (unless somebody thinks of an improvement)

I also run my tests like you do it : 
```bash
cd src/
pytest
```

It's odd, but no more problem if you removed the code ;)

danielquinn:
Sweet!  Thanks for your work on this :-)

PR #328, opened 2018-03-23
Use --noreload for permanent server start commands

erikarvstedt:
Without it, the server is highly resource-intensive even when running idle.
On my system, it permanently consumed 30% of one core.

I've rephrased the paragraph starting with "Note that" in `utilities.rst` to avoid repeating the word "continuously".


danielquinn:
I didn't know there was a way to skip the source-checking.  Thanks!

PR #333, opened 2018-03-29
Fix runtime warning when adding log entries

erikarvstedt:
`LogEntry.action_time` expects a Django timezone object instead of a builtin datetime.

This fixes a runtime warning of the following kind:
```
RuntimeWarning: DateTimeField LogEntry.action_time received a naive datetime (2018-03-21 20:51:01.714123) while time zone support is active.
```


danielquinn:
Ooh, thanks for finding & fixing this one!

PR #343, opened 2018-04-26
Add field 'added' to documents

erikarvstedt:
This field indicates when the document was added to the database.

Although this field is not yet exposed to the UI, I think it's very useful to store this information.

### DIFF
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -125,7 +125,9 @@ class DocumentAdmin(CommonAdmin):
         }
 
     search_fields = ("correspondent__name", "title", "content", "tags__name")
-    list_display = ("title", "created", "thumbnail", "correspondent", "tags_")
+    readonly_fields = ("added",)
+    list_display = ("title", "created", "added", "thumbnail", "correspondent",
+                    "tags_")
     list_filter = ("tags", "correspondent", FinancialYearFilter,
                    MonthListFilter)
 
--- a/src/documents/migrations/0020_document_added.py
+++ b/src/documents/migrations/0020_document_added.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.utils.timezone
+
+
+def set_added_time_to_created_time(apps, schema_editor):
+    Document = apps.get_model("documents", "Document")
+    for doc in Document.objects.all():
+        doc.added = doc.created
+        doc.save()
+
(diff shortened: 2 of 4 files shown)


danielquinn:
This is a good idea, but if I could make a request as to how you've implemented this, I'd prefer that we not allow for null values if we can help it.  It tends to present a hazard to things like sorting, date math, etc.

Instead, let's drop `auto_now_add=True` in favour of `default=timezone.now` and drop the `null=` and `blank=` references as well.

Finally, a migration that would set all past `added` values to whatever is currently in `created` (or `modified`?  What do you think?) would help users with existing setups have a value to work with (even if it's not necessarily the precise time the document was added.

Note that to make this work, you'll probably have to set a temporary default for the first migration (Django will likely require it).  The easiest thing I've seen for this is to set a date in the distant future (say, `2020-01-01`) and then write your migration to just overwrite all `added` values as discussed above.

erikarvstedt:
These are great suggestions, thanks!
Unfortunately, I've never worked with Django and I don't know how to tackle the migration.
To every Django dev who'd want to help: Feel free to implement and commit this feature by yourself; there's no need to credit me.

erikarvstedt:
I've updated the admin view to include the new field.

@danielquinn Could you help me with the migration? I just need a rough sketch and I'll figure out the rest.

danielquinn:
Hi @erikarvstedt, here's what needs to be done to get the migrations working.  Basically it's 3 steps:

1. Create a migration that adds the `added` column, and sets each value to the assigned default: `timezone.now()`.
2. Write a data migration to manually set the `added` date to the `modified` value of each document.

### The first migration

Edit the `added` definition to look like this:

```python
added = models.DateTimeField(
    default=timezone.now, editable=False, db_index=True)
```

Then create a migration for it with `./manage.py makemigrations`.

### The data migration

A *data migration* is the official way to run arbitrary Python code to move data around during a migration process.  The process for that is simple and [the official documentation has you covered](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations).  Here's the short version though:

1. `./manage.py makemigrations documents --empty`
2. Open the created file and add add your arbitrary code.  This code should just loop over all the documents and set `document.added = document.modified`.    The Django documentation should giv eyou everything you need to get started, but if you have more questions, feel free to ask here.

erikarvstedt:
All done, thanks for the help!

danielquinn:
Hooray!  Thanks for your work on this.  I just tested out the new feature and it works swimmingly.

PR #344, opened 2018-04-26
[Help needed] Increase link area in documents listing

### LINKED ISSUE #335: UI: Increase clickable area to enter document edit mode
Currently, only the document title text is clickable, which is frustrating for short document titles like single digits.
Increase the clickable area to include the whole visual header.

erikarvstedt:
Increase the link area to include the whole visual header.

Fixes #335 

This is not an ideal solution because it 
1. requires Javascript
2. doesn't endow the header with link semantics, so, for example, the browser won't show a link preview on mouseover.

Maybe someone well-versed in CSS can (later) figure out a better approach.

Edit:
I've just added `WIP` to the title as I noticed that the `onclick` handler is erroneously triggered when the checkbox is clicked.
Is there a web dev out there who can help me finish this PR? :disappointed_relieved: :wink:

### DIFF
--- a/src/documents/templates/admin/documents/document/change_list_results.html
+++ b/src/documents/templates/admin/documents/document/change_list_results.html
@@ -132,9 +132,10 @@
     {# 3: Image #}
     {# 4: Correspondent #}
     {# 5: Tags #}
+    {# 6: Document edit url #}
     <div class="box">
       <div class="result">
-        <div class="header">
+        <div class="header" onclick="location.href='{{ result.6 }}';" style="cursor: pointer;">
           <div class="checkbox">{{ result.0 }}</div>
           <div class="info">
             {{ result.4 }}<br />
--- a/src/documents/templatetags/hacks.py
+++ b/src/documents/templatetags/hacks.py
@@ -1,3 +1,5 @@
+import re
+
 from django.contrib.admin.templatetags.admin_list import (
     result_headers,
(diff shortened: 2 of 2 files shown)


danielquinn:
The hack you added to the Python part is a special bit of beautiful hackery.  I love it.  It's hack in a file already full of hacks, so I have to hate it, but I still love it.  Just please put constants like `EXTRACT_URL` at the top of the file (after imports, before functions) just for consistency of style.

As for front-end stuff, I'm afraid I suck at that stuff (but you know that).  You might have better luck wrapping the text inside the `.info` block with a link of its own and styling that link to have a fixed `height` / `width` or maybe fiddling with a `padding:` value.

erikarvstedt:
I've moved the constant definition.
Since I'm writing very little Python day-to-day, even small hints about style and idioms are super helpful. Please never hesitate to correct me!

danielquinn:
I've had a lot on my plate these last few weeks, and I'm finally getting around to Paperless again.  Thanks for your patience :-)

erikarvstedt:
@danielquinn , thanks for picking up working on Paperless again, it's highly appreciated! :heart:
Alas, the merge was a bit premature, since the PR wasn't yet finished (the `onclick` handler still triggers when clicking the checkbox.)
I'll dig into CSS and figure out how to do this properly. I'll post a fix tomorrow.

PR #351, opened 2018-05-11
Consumer improvements, inotify support

erikarvstedt:
Please see the individual commit messages for in-depth explanations of all changes.

I've been using inotify in production for a few weeks now and it all works perfectly.

### DIFF
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -48,7 +48,7 @@ class Consumer:
         self.consume = consume
         self.scratch = scratch
 
-        make_dirs(self.scratch)
+        os.makedirs(self.scratch, exists_ok=True)
 
         if not self.consume:
             raise ConsumerError(
@@ -247,11 +247,3 @@ class Consumer:
         with open(doc, "rb") as f:
             checksum = hashlib.md5(f.read()).hexdigest()
         return Document.objects.filter(checksum=checksum).exists()
-
-
-def make_dirs(*dirs):
-    for dir in dirs:
-        try:
-            os.makedirs(dir)
-        except FileExistsError:
-            pass
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -75,80 +75,82 @@ class Consumer:
         docs_old_to_new = sorted(docs, key=lambda doc: os.path.getmtime(doc))
 
         for doc in docs_old_to_new:
+            self.try_consume_file(doc)
 
-            doc = os.path.join(self.consume, doc)
+    def try_consume_file(self, doc):
+        doc = os.path.join(self.consume, doc)
 
-            if not os.path.isfile(doc):
-                continue
+        if not os.path.isfile(doc):
+            return
 
-            if not re.match(FileInfo.REGEXES["title"], doc):
-                continue
+        if not re.match(FileInfo.REGEXES["title"], doc):
+            return
 
-            if doc in self._ignore:
-                continue
+        if doc in self._ignore:
+            return
 
-            if not self._is_ready(doc):
-                continue
+        if not self._is_ready(doc):
+            return
 
-            if self._is_duplicate(doc):
-                self.log(
-                    "info",
-                    "Skipping {} as it appears to be a duplicate".format(doc)
-                )
-                self._ignore.append(doc)
-                continue
+        if self._is_duplicate(doc):
+            self.log(
+                "info",
+                "Skipping {} as it appears to be a duplicate".format(doc)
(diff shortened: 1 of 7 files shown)


src/documents/mail.py
```diff
@@ -42,7 +42,7 @@ class Message(Loggable):
     and n attachments, and that we don't care about the message body.
     """
 
-    SECRET = os.getenv("PAPERLESS_EMAIL_SECRET")
+    SECRET = os.getenv("PAPERLESS_EMAIL_SECRET", "")
```
danielquinn: This is probably a Bad Idea, as testing for `""` in any string will always return `True`, negating the value of the secret in the first place.  I figure you did this 'cause later down we test `if self.SECRET not in self.body` and `None` in this case will break things, but the better way to account for this would be to have a test in `__init__()` for whether `self.SECRET is None` and raise an exception instead.

src/documents/mail.py
```diff
@@ -75,8 +75,9 @@ def __init__(self, data, group=None):
                 continue
 
             dispositions = content_disposition.strip().split(";")
-            if not dispositions[0].lower() == "attachment" and \
-               "filename" not in dispositions[1].lower():
+            if len(dispositions) < 2 or \
```
danielquinn: 3-line if statements make me cringe a little.  I didn't much like it when it was 2-lines, but at 3 we're pushing it.  Can you maybe break this up into multiple if statements just for readability?

src/documents/consumer.py
```diff
@@ -224,24 +242,16 @@ def _cleanup_doc(self, doc):
...
     def _is_duplicate(doc):
         with open(doc, "rb") as f:
             checksum = hashlib.md5(f.read()).hexdigest()
         return Document.objects.filter(checksum=checksum).exists()
+
+
+def make_dirs(*dirs):
+    for dir in dirs:
+        try:
+            os.makedirs(dir)
```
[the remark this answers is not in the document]
erikarvstedt: Yeah, with this kwarg we don't need the helper fn.

src/documents/mail.py
erikarvstedt: I actually did this on purpose.
For use cases where no email secret is needed the empty string is an elegant way to disable it. 
Why should a secret be mandatory?
(It's not a secret by the way, it's more like a marker, but I don't think it's worth renaming it.)

src/documents/mail.py
erikarvstedt: Done.

erikarvstedt:
When all looks good to you I'll apply the fixup commits and we're ready to merge!

src/documents/mail.py
danielquinn: That's the thing, it's meant to be a secret.  It prevents other people from stuffing your server with bogus (or even potentially nefarious) documents.

For example, if I know your target email address, I can send a bunch of PDF documents to your server and do any or all of the following:

1. Fill up your hard drive (best case)
2. Use a 0-day exploit that triggers something scary on your server.
3. Use an anonymous email account to fill your personal document store with child porn.

...all by just emailing you something

The secret was added to make sure that emails are from a source you know you can safely attempt to import.  It doesn't have to be an unmemorable string like `cae3Mai1xi?uSood`, but even insisting that every email must contain `parse-all-the-things` will protect you against `a`, `b`, and `c` above.

erikarvstedt: I envisioned a use case where the email address **is** the secret.
But yeah, most people won't set it up that way and it may be better to nudge them towards a safer configuration.
Fixed.

danielquinn:
So it turns out that the argument is `exist_ok=` and not `exists_ok=`.  My bad, sorry.  I fixed the kwarg and merged your changes.  Thank you so much!  I figured there was a way to do this better, but I didn't know how when I first wrote this thing.

erikarvstedt:
Those fixup commits were meant to be squashed to avoid polluting the history.
If you don't mind the `git push --force`, we could still fix this. Would that be ok?

danielquinn:
Normally, I'd be cool with that, but as a lot of people are running instances by checking out master, this could confuse/upset an indeterminite number of people.  I'm less concerned about history (we've got merge commits anyway) and more concerned with user experience.

You seem to be a lot more adept at git than I am though, so advice for future stuff is appreciated :-)

ovv:
My opinion is to not do it :)

erikarvstedt:
Alright, thanks for reviewing and the helpful suggestions! :relaxed:

PR #352, opened 2018-05-15
fix bug where docker-entrypoint.sh exits w/o notice

Strubbl:
This commit fixes a nasty bug, where the docker-entrypoint.sh silently
exits without any error message. The test for a lock file can fail and
due to the `set -e` at the beginning of the file the bash script exists
without starting the paperless application.
It is fixed by moving the check for the existence of the lock file into
the if statement, where the `set -e` does not trigger an exit in case
the statement fails (see https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html in the section for parameter `-e`).

Additionally this commit enables the script to trap exit signals and in
that case deletes the lock file.

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -46,11 +46,10 @@ migrations() {
     # A simple lock file in case other containers use this startup
     LOCKFILE="/usr/src/paperless/data/db.sqlite3.migration"
 
-    set -o noclobber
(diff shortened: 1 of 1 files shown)


danielquinn:
The force is strong with your Bash skills friend.  That's a level of kung-foo well beyond me.  Thanks for finding & fixing this!

Issue #253, opened 2017-08-10
Deploye at non root URL


viharm:
This seems like a very good application. Thank you for efforts in creating this app. I respect the vision and agree with it.

I am trying to deploy this behind in existing environment which has an *Apache* installation serving several websites.

I would like to use my *Apache* webserver as a reverse proxy to access *paperless*. However I am also deploying this at a non-root URL of an existing virtual host, e.g., *https://mydomain.tld/paperless*

When I setup the reverse proxy, the Django service keeps throwing back the root URL, thus failing. Is there a configuration for setting this path/URL?

Regards.

danielquinn:
I've *never* done something like that before, but a cursory search online pointed me to `[FORCE_SCRIPT_NAME](https://docs.djangoproject.com/en/1.10/ref/settings/#force-script-name)` which might be your best bet.

I suggest editing `paperless/settings.py` and adding `FORCE_SCRIPT_NAME = "paperless"` to the bottom and restarting the server to see if that works.  If it does, issue a pull request that follows the other settings vars use of environment variables and I'll roll it into master.

If it doesn't work however, I'm afraid I don't know how to do that :-(  I'm not opposed to making it work though, should someone else know how to do it.

MasterofJOKers:
I only know how to do this for deployments using uWSGI. This is part of my nginx config for another project:
```
location /blog/ {
                uwsgi_pass      127.0.0.1:8780;
                include         uwsgi_params;
                uwsgi_param SCRIPT_NAME /blog;
                # tell uwsgi to remove SCRIPT_NAME from the URL
                uwsgi_modifier1 30;
        }
```
Googling for SCRIPT_NAME, it seems also possible to set this as environment variable for django in a wsgi file. But that setting sounds like it should do the job.

kmlucy:
I run my instance at /paperless/. I run in docker, so I just run this before it starts to modify the necessary hardcoded URLs:
```
sed -i 's=/fetch=/paperless/fetch=' /usr/src/paperless/src/documents/admin.py
sed -i -e 's=/admin/login=/paperless/admin/login=' -e 's=/static/=/paperless/static/=' -e 's=/media/=/paperless/media/=' /usr/src/paperless/src/paperless/settings.py
sed -i -e 's=r"^=r"^paperless/=' -e 's=r"admin=r"paperless/admin=' -e 's=r""=r"paperless"=' /usr/src/paperless/src/paperless/urls.py
```

It's not the cleanest way, but it does work.

[...]

PR #353, opened 2018-05-15
[WIP] Docker use gunicorn deploy

Strubbl:
The docker container uses gunicorn to launch the application. Static files need to be served separately.

So i am using paperless in a subpath of my domain now. Maybe #253 is an related issue?

This is the snippet of my nginx.conf:

```
        location /pl/static/ {
            alias /data/paperless/static/;
        }
        location /pl {
            proxy_set_header Host $host;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_pass http://127.0.0.1:8000;
        }
```

This is my create call for the webserver docker:

`docker create --name $CONTAINERNAME -p 8000:8000 -v /data/paperless/data:/usr/src/paperless/data -v /data/paperless/media:/usr/src/paperless/media -v /data/paperless/static:/usr/src/paperless/static -e PAPERLESS_PASSPHRASE=supersecret -e PAPERLESS_FORCE_SCRIPT_NAME=/pl -e PAPERLESS_STATIC_URL=/pl/static/ strubbl_paperless gunicorn`

I am not familiar with and not using docker-compose and therefore did not update any related files. Maybe someone can help here?

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -59,6 +59,7 @@ initialize() {
     map_uidgid
     set_permissions
     migrations
+    sudo -HEu paperless "/usr/src/paperless/src/manage.py" "collectstatic" "--clear" "--no-input"
 }
 
 install_languages() {
@@ -97,6 +98,11 @@ if [[ "$1" != "/"* ]]; then
     # Install additional languages if specified
     if [ ! -z "$PAPERLESS_OCR_LANGUAGES"  ]; then
         install_languages "$PAPERLESS_OCR_LANGUAGES"
+    fi
+
+    if [ "$@" = "gunicorn" ]; then
+      cd /usr/src/paperless/src/ && \
+      exec sudo -HEu paperless GUNICORN_CMD_ARGS="--bind=0.0.0.0:8000 --access-logfile=- --log-file=-" /usr/bin/gunicorn paperless.wsgi
     fi
 
     exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -55,11 +55,15 @@ migrations() {
     fi
 }
 
+collectstatic() {
+    sudo -HEu paperless "/usr/src/paperless/src/manage.py" "collectstatic" "--clear" "--no-input"
+}
+
 initialize() {
     map_uidgid
     set_permissions
     migrations
-    sudo -HEu paperless "/usr/src/paperless/src/manage.py" "collectstatic" "--clear" "--no-input"
+    collectstatic
 }
 
 install_languages() {
@@ -100,12 +104,15 @@ if [[ "$1" != "/"* ]]; then
         install_languages "$PAPERLESS_OCR_LANGUAGES"
     fi
 
-    if [ "$@" = "gunicorn" ]; then
-      cd /usr/src/paperless/src/ && \
-      exec sudo -HEu paperless GUNICORN_CMD_ARGS="--bind=0.0.0.0:8000 --access-logfile=- --log-file=-" /usr/bin/gunicorn paperless.wsgi
-    fi
+    if [[ "$1" = "gunicorn" ]]; then
+        supplied_args="${@:2}"
+        gunicorn_cmd_args="${supplied_args:---bind=0.0.0.0:8000 --access-logfile=- --log-file=-}"
 
-    exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
+        cd /usr/src/paperless/src/ && \
+            exec sudo -HEu paperless GUNICORN_CMD_ARGS="$gunicorn_cmd_args" /usr/bin/gunicorn paperless.wsgi
+    else
+        exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
+    fi
 fi
 
 exec "$@"
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN apk update --no-cache && apk add python3 gnupg libmagic libpq bash shadow cu
 
 WORKDIR /usr/src/paperless/src
 # Mount volumes and set Entrypoint
-VOLUME ["/usr/src/paperless/data", "/usr/src/paperless/media", "/consume", "/export"]
+VOLUME ["/usr/src/paperless/data", "/usr/src/paperless/media", "/consume", "/export", "/usr/src/paperless/static"]
 ENTRYPOINT ["/sbin/docker-entrypoint.sh"]
 CMD ["--help"]
 
--- a/src/paperless/urls.py
+++ b/src/paperless/urls.py
@@ -36,7 +36,7 @@ urlpatterns = [
 
     # File downloads
     url(
-        r"^fetch/(?P<kind>doc|thumb)/(?P<pk>\d+)$",
+        r"fetch/(?P<kind>doc|thumb)/(?P<pk>\d+)$",
         FetchView.as_view(),
         name="fetch"
     ),


scripts/docker-entrypoint.sh
```diff
@@ -60,6 +60,7 @@ initialize() {
     map_uidgid
     set_permissions
     migrations
+    sudo -HEu paperless "/usr/src/paperless/src/manage.py" "collectstatic" "--clear" "--no-input"
```
pitkley: Nit: Even though it is only one line, I'd feel it be better placed in its own function, similar to `migrations`.

scripts/docker-entrypoint.sh
```diff
@@ -100,6 +101,11 @@ if [[ "$1" != "/"* ]]; then
         install_languages "$PAPERLESS_OCR_LANGUAGES"
     fi
 
+    if [ "$@" = "gunicorn" ]; then
+      cd /usr/src/paperless/src/ && \
+      exec sudo -HEu paperless GUNICORN_CMD_ARGS="--bind=0.0.0.0:8000 --access-logfile=- --log-file=-" /usr/bin/gunicorn paperless.wsgi
+    fi
```
pitkley: Even though I won't be needing it, I think that the args passed to gunicorn should be configurable. Maybe use something like the following instead?

```bash
if [[ "$1" = "gunicorn" ]]; then
    supplied_args="${@:2}"
    gunicorn_cmd_args="${supplied_args:---bind=0.0.0.0:8000 --access-logfile=- --log-files=-}"

    cd /usr/src/paperless/src/ && \
        exec sudo -HEu paperless GUNICORN_CMD_ARGS="$gunicorn_cmd_args" /usr/bin/gunicorn paperless.wsgi
else
    exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
fi
```

This will check if the first supplied parameter is `gunicorn`, and will pass the rest into `GUNICORN_CMD_ARGS` if specified, otherwise the defaults you had shown.

(I have also added the `else`-branch, because I find this more specific, although it doesn't change the behaviour at all.)

pitkley (approved):
I have some comments/ideas, nothing major though.

As far as I can tell, this should not affect the existing compose-specifications at all. They simply won't use what is enabled through this change.

addadi:
looks good to me

danielquinn:
@Strubbl what do you think of @pitkley's change suggestions?

Strubbl:
I appreciate his feedback and implemented it, but i pushed to early. I am just getting an error. I report back when i fully tested the change.

Strubbl:
@danielquinn There is another open point for hosting paperless in a subpath, which i have missed because i was always logged in. Currently, when i am not logged in and i surf to /subpath i get redirected to /admin and not to /subpath/admin/
Can you please help me, how i can easily redirect /subpath to /subpath/admin. I think we need to change the line
https://github.com/danielquinn/paperless/blob/master/src/paperless/urls.py#L48
But i do not know how to reuse the environment variable or the setting.

danielquinn:
Hi @Strubbl, I think I've fixed this in master now, so if you do a pull, it *should* work for you and everyone else.

Basically I've removed the hard-coded reference to `/admin/` in the redirect and instead am now using `reverse_lazy("admin:index")` which should take `FORCE_SCRIPT_NAME` into account when returning the URL.  Non-subpath people shouldn't be affected.

Give that a shot, and let me know how it pans out for you.

Strubbl:
Unfortunately, this did not fix the redirect yet. There is still a bug in the redirect. But this was already present before your commit. Viewing from user perspective the commit changed nothing. It is still the same behaviour:

I call the url, where i expect paperless: 
http://asdf.de/pl/

and get the error message:
```
Page not found (404)
Request Method: 	GET
Request URL: 	http://asdf.de/pl/pl/

Using the URLconf defined in paperless.urls, Django tried these URL patterns, in this order:

    ^api/auth/
    ^api/
    fetch/(?P<kind>doc|thumb)/(?P<pk>\d+)$ [name='fetch']
    ^push$ [name='push']
    admin/
    ^$
    ^media\/(?P<path>.*)$

The current path, pl/, didn't match any of these.

You're seeing this error because you have DEBUG = True in your Django settings file. Change that to False, and Django will display a standard 404 page.
```

The gunicorn log says:
```
Not Found: /pl/pl
172.17.0.1 - strubbl [28/May/2018:17:32:03 +0000] "GET /pl HTTP/1.0" 404 2668 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0"
```

Here is a double "pl/pl". But why? i don't get it.

Funnily enough, the path to admin still works as expected:
`172.17.0.1 - strubbl [28/May/2018:17:32:55 +0000] "GET /pl/admin/ HTTP/1.0" 200 8028 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0"`

danielquinn:
To my knowledge, support for Paperless on a subpath has never been very good, but maybe we can change that. Can you share your paperless.conf (or docker-compose.env) file?  Obviously, you can strip the private stuff like passphrase etc.  I'll try to find some time to reproduce this locally if I can get Nginx to play with it.

Strubbl:
My docker/paperless configuration is the docker create command in the first post https://github.com/danielquinn/paperless/pull/353#issue-188204598
This has not changed yet.

I did not configure a file called paperless.conf. So except for what i am overriding by my env. variables with the docker create command, it should be "all default". 

I create the image with:

1. cd paperless
2. docker build --pull -t strubbl_paperless .
3. $DOCKER_CREATE_COMMAND_FROM_ABOVE where $CONTAINERNAME equals strubbl-paperless-webserver
4. docker start strubbl-paperless-webserver

danielquinn:
I'm sorry.  I just don't know enough about the various required Docker incantations to make any sense of this.  When I run:

```
$ docker run --name strubbl-paperless-webserver -p 8000:8000 -v /home/daniel/Projects/Paperless/data:/usr/src/paperless/data -v /home/daniel/Projects/Paperless/media:/usr/src/paperless/media -v /home/daniel/Projects/Paperless/static:/usr/src/paperless/static -e PAPERLESS_PASSPHRASE=supersecret -e PAPERLESS_FORCE_SCRIPT_NAME=/pl -e PAPERLESS_STATIC_URL=/pl/static/ strubbl_paperless gunicorn
```

It just hangs there doing nothing and sending a `curl` request to `localhost:8000` fails with "connection reset".

Strubbl:
Do you get any error message?

I am getting output like this and the command line is still blocked by gunicorn awaiting/serving requests:

```
…
Copying '/usr/lib/python3.6/site-packages/rest_framework/static/rest_framework/js/jquery-3.3.1.min.js'
Copying '/usr/lib/python3.6/site-packages/rest_framework/static/rest_framework/js/prettify-min.js'

108 static files copied to '/usr/src/paperless/static'.
[2018-07-08 14:38:05 +0000] [1] [INFO] Starting gunicorn 19.8.1
[2018-07-08 14:38:05 +0000] [1] [INFO] Listening at: http://0.0.0.0:8000 (1)
[2018-07-08 14:38:05 +0000] [1] [INFO] Using worker: sync
[2018-07-08 14:38:05 +0000] [42] [INFO] Booting worker with pid: 42
```

I rebased my commits to current master HEAD to make the test you did. 

Of course when i surf to localhost:8000 i see paperless only without the assets (404 errors) as they have been configured to be served via nginx.

danielquinn:
Ok I finally got it working -- I'm not sure what I was doing wrong the first few times, but now I see what you see: a working Paperless installation without static files.

The static URLs are a little weird though, as you say: I'm seeing a lot of `/pl/pl/`.  I noted that you've set `PAPERLESS_FORCE_SCRIPT_NAME=/pl` and `PAPERLESS_STATIC_URL=/pl/static/`, but on a hunch I restarted the container with `PAPERLESS_STATIC_URL=/static` (note the missing `/pl`) and the URLs are showing up as I'd expect.  I'm thinking that Django's `STATIC_URL` is meant to be mounted *on top* of whatever `FORCE_SCRIPT_NAME` is set to, so when they both are prefixed with `/pl`, that weirdness happens.

Strubbl:
I just updated this gunicorn PR branch to rebase on current master HEAD commit. It still works.

One issue i see is that the URL https://paperless/pl is a 404. It does not get redirected to /pl/admin. Full error message is: 

```
Using the URLconf defined in paperless.urls, Django tried these URL patterns, in this order:

    ^api/auth/
    ^api/
    fetch/(?P<kind>doc|thumb)/(?P<pk>\d+)$ [name='fetch']
    ^push$ [name='push']
    admin/
    ^$
    ^media\/(?P<path>.*)$

The current path, pl, didn't match any of these.
```


Then i go to the URL https://paperless/pl/admin and i get redirected to https://paperless/pl/admin/login/?next=/pl/pl/admin/.
One can already see the double "pl/" in the URL. But i ignore that for now and do login. After login i get redirected to the double pl URL: https://paperless/pl/pl/admin/
This is of course ugly.

To workaround this ugly URL, i can type in https://paperless/pl/admin/ and i get the same page with the nice URL. I am still logged in.


Currently i do not know how to fix this wrong redirecting issue and the 404 issue. --> So i do not know how to proceed with this PR. If it is okay for you, i would leave it open until someone™ (maybe me) finds a fix.

pitkley:
@Strubbl what do you think, did PR #564 cover what you intened to change with your PR? Are the URL-changes or anything else still necessary?

PR #358, opened 2018-05-30
Fix unclickable checkbox in documents view

erikarvstedt:
1. Clicks to the document selection checkbox were captured by the onclick
   handler of the document item header. This is now fixed.

2. Reexpose the doc title link to mouse events by putting it on top of
   the header link layer.

### DIFF
--- a/src/documents/templates/admin/documents/document/change_list_results.html
+++ b/src/documents/templates/admin/documents/document/change_list_results.html
@@ -29,13 +29,32 @@
   .result .header {
     padding: 5px;
     background-color: #79AEC8;
+    position: relative;
   }
-  .result .header .checkbox{
+  .result .header .checkbox {
     width: 5%;
     float: left;
+    position: absolute;
+    z-index: 2;
   }
   .result .header .info {
     margin-left: 10%;
+    position: relative;
+  }
+  .headerLink {
+    cursor: pointer;
+    opacity: 0;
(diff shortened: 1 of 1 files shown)


erikarvstedt:
Ready to merge. :rocket:

danielquinn:
I have some concerns about the accessibility of depending on javascript-only for this link, so I've asked a friend who knows a lot more about this than I do to have a look before I merge it.  I'll let you know what comes out of it.

erikarvstedt:
Note that the original title text link is still available in any case, JS is only used to extend the clickable area to the whole header background.
At first I used an `<a>` spanning the whole header background, but that made it impossible to select the title and correspondent text (tested in Firefox and Chrome). I find that much worse for accessibility than the JS-only background link.

I'd be excited to hear if there's a better solution.

erikarvstedt:
For now I think we should use this solution because it fixes a real bug.
Also, the accessability/usability is in no way diminished when compared to the original Django behaviour, even when JS is disabled.

danielquinn:
Alright I'll test this out locally after work today then.

danielquinn:
Apologies.  After checking out the code and running it locally, I realise what you meant by the `<a href=""` bit still being there to work with.  I couldn't see it in the pull request, so naturally, I was confused about what you were talking about.  I hadn't realised that Django was generating the link itself.

Issue #362, opened 2018-06-13
GnuPG changed default environment variables (initial migrate broken)


RNCTX:
I'm on FreeBSD-11.1-RELEASE, not sure about Linux distros.

```
$> gpg --version
gpg (GnuPG) 2.2.8
libgcrypt 1.8.2
Copyright (C) 2018 Free Software Foundation, Inc.

```
gnupghome doesn't work anymore

homedir is the new equivalent

On a personal note, changing variable names to those that are somehow more explicit or descriptive might be an issue worth debating for people who would rather debate on mailing lists than do anything else.  Changing it to one that's LESS descriptive and LESS explicit than it was before is just being more Linux than Linux people usually even are.  

Going by [this](https://stackoverflow.com/questions/35028852/how-to-set-the-gnupg-home-directory-within-the-gnupg-python-binding) the change would have happened in/before 2016?  Not sure, haven't researched it.  The attached pull fixes it.

RNCTX:
Whoops, multiple account juggling foolery, lets try that again...

PR #364, opened 2018-06-13
Fix GnuPG variable change from "gnupghome" to "homedir"

RNCTX:
#362

### DIFF
--- a/src/documents/migrations/0012_auto_20160305_0040.py
+++ b/src/documents/migrations/0012_auto_20160305_0040.py
@@ -19,7 +19,7 @@ class GnuPG(object):
     A handy singleton to use when handling encrypted files.
     """
 
-    gpg = gnupg.GPG(gnupghome=settings.GNUPG_HOME)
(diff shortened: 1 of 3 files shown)


ddddavidmartin:
I'm not familiar with gnupg, but it looks like this may need a version check or some handling to allow for both variants. The travis-ci builds are falling over for example.

RNCTX:
Aha! 

> This is a modified version of python-gnupg, (forked from version 0.3.2) which was created by Vinay Sajip, which itself is a modification of GPG.py written by Steve Traugott, which in turn is a modification of the pycrypto GnuPG interface written by A.M. Kuchling. This version is patched to sanitize untrusted inputs, due to the necessity of executing subprocess.Popen([...], shell=True) in order to communicate with GnuPG. Several speed improvements were also made based on code profiling, and the API has been cleaned up to support an easier, more Pythonic, interaction.

I’ll have to double check later but it might be because I’m using python 3.6 and defaulting over to a compatible fork.

RNCTX:
Confirmed, conflicting gnupg python libs on my end.  Closing.

PR #369, opened 2018-06-18
Update docs

HerrSubset:
The guesswork section of the docs did not explicitly mention the `Date - Correspondent - Title.pdf` format. From my experience, this format works however. I added it for clarity.

I didn't check the source code, so the support for this format might be unintentional.

### DIFF
--- a/docs/guesswork.rst
+++ b/docs/guesswork.rst
@@ -20,6 +20,8 @@ for you.  This is is the logic the consumer follows:
    the pattern: ``Date - Correspondent - Title - tag,tag,tag.pdf``.  Note that
    the format of the date is **rigidly defined** as ``YYYYMMDDHHMMSSZ`` or
    ``YYYYMMDDZ``.  The ``Z`` refers "Zulu time" AKA "UTC".
+   The tags are optional, so the format ``Date - Correspondent - Title.pdf``
+   works as well.
 2. If that doesn't work, we skip the date and try this pattern:
    ``Correspondent - Title - tag,tag,tag.pdf``.
 3. If that doesn't work, we try to find the correspondent and title in the file


danielquinn:
I always love it when a new user tweaks the docs for clarity.  Thank you so much!

HerrSubset:
You're welcome, I really like what you guys made so far.

I'm still working on a flow to get all my files into paperless, but maybe I'll contribute some code as well in the future :)

Strubbl:
What happens after paperless consumption of that documents? Do you forget about them and find documents later on via search? Or are you running some manual organization steps afterwards?

I think my flow is currently way too complicated and i need to think about if i am doing too much to keep my files organized.

1. Scan documents as JPEGs to SD card
2. Copy documents from SD to PC
3. Group JPEGs into folders cause most documents consist of more than one page
4. Rename folder according to guesswork format to include date, correspondant, title, tags (this is the most time consuming step. Of course most tags are auto-applied by paperless, but some i add manually)
5. Run simple script, which creates PDF, optimizes it for file size and copies it to paperless consume dir

Maybe my fourth step is too much in terms of wasted time for documents you'll never look into again. :thinking:

danielquinn:
Wow.  Yeah that's way too much work for me :-)

* I've eliminated steps 1 and 2 by making sure my scanner can push to FTP, so you might wanna see if that's an option for you.
* My scanner also will create multi-page documents, so your steps 3 & 5 aren't a problem for me.
* Importantly though, your step 4 is just not something I do.  My scanner creates a file called something like `______<random-number>.pdf` and that's what Paperless consumes.  Then, it uses the auto-tagging and auto-correspondent setting signals in Paperless to at least try to set the correspondent and tags.  For document title, I occasionally log into Paperless and try to set these manually.

For the most part, I follow the "Google methodology" of "just make everything searchable".  So I've got a lot of documents in Paperless that haven't been assigned proper titles, but have a guessed correspondent and a list of attached tags.  That, coupled with the complete indexing of the document contents is usually enough for me to find what I want.

For example, when I was filing my visa application recently, I just typed "easyjet" and got all of my flights in and out of the UK over the last few years.  "national insurance" was enough to bring up the relevant government documents I needed too.  I suppose I *could* go over it all and organise it neatly with proper names, but that's too much work ;-)

HerrSubset:
I'd use the same workflow, @danielquinn if it weren't for my scanner not supporting FTP.

So my workflow looks something like this:

1. Use some scanning app to scan documents straight to pdf.
2. From the app, upload the scan to dropbox.
2. Bulk-upload the files from dropbox to paperless. Using the naming conventions, I can already set the correspondent, date,  tags,... 

But maybe setting these things is not even necessary. Especially for downloaded pdfs, where paperless can just copy the text. But it should allow me to ditch the FTP server. It can upload everything straight to the paperless API.

PR #374, opened 2018-07-04
add search field on admin start page and module start page

elohmeier:
99% of the time when I open the Paperless web interface I want to quickly search for a document using the search field in the documents list.
Currently this requires users to open the documents list, wait for their connection to transfer a lot of not relevant thumbnail images, then type in the search term and then hit search.

To make Paperless a bit more mobile-friendly and to improve the workflow a bit for this common scenario, I've added a search box to the start page. This search box is referencing the default document_changelist search functionality (checkout the source code).
I've tried to implement this change with the minimal modifications needed, only by adding a form in the HTML.

This is how it looks on the desktop:
image

This is how it looks on mobile:
image

After long consideration I've recently switched from (the also great) Mayan EDMS to Paperless and am fully delighted - thanks for this great piece of software! I can now run my DMS on a much smaller and more resource-saving machine and modifications are a breeze.

### DIFF
--- a/src/documents/templates/admin/index.html
+++ b/src/documents/templates/admin/index.html
@@ -4,6 +4,24 @@
 {% load i18n static %}
 
 
+{# This block adds a search form on the admin start page and on the module start page so that #}
+{# the user can quickly search for documents #}
+{% block pretitle %}
+<div>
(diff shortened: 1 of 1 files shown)


danielquinn:
This is a great addition!  Thanks so much :-)

PR #375, opened 2018-07-05
add document count column for tags and correspondents

elohmeier:
This adds a column to the "Tags" and "Correspondents" view.
It's useful for finding unused tags or correspondents.

Screenshot:
image


danielquinn:
Another nice & simple addition.  Thanks again!

PR #376, opened 2018-07-05
optimize Dockerfile for caching

elohmeier:
I've reordered the Dockerfile a bit. This helps users who change stuff in the Django-app and want to rebuild it using ```docker-compose up -d --build```.
Docker is then noticing that e.g. the requirements.txt didn't change and that no package reinstallation is necessary. It'll quickly jump to the step where the application files are added, speeding up the development process for Docker users.

See the screenshot of a cached build for clarification:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,8 @@ LABEL maintainer="The Paperless Project https://github.com/danielquinn/paperless
       contributors="Guy Addadi <addadi@gmail.com>, Pit Kleyersburg <pitkley@googlemail.com>, \
         Sven Fischer <git-dev@linux4tw.de>"
 
-# Copy application
+# Copy requirements file and init script
 COPY requirements.txt /usr/src/paperless/
-COPY src/ /usr/src/paperless/src/
-COPY data/ /usr/src/paperless/data/
-COPY media/ /usr/src/paperless/media/
 COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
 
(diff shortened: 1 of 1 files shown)


danielquinn:
Neat trick!  Honestly, I look at your changes and I don't understand *how* it works, but I've just tested this and sure enough, the build process is drastically reduced.  I'll merge this now, but would you mind further explaining just what's going on here?  I was under the impression that `ENTRYPOINT` and `CMD` had to be at the bottom of the file, so moving the `COPY` lines below them seems counterintuitive.

elohmeier:
To explain what's going on: When building, Docker will keep track of the files added with each build step in it's cache. It's storing checksums for each file to skip build steps when they probably don't need to be repeated. E.g. if you would add a python package to the requirements.txt, Docker would notice the invalidated cache because of that checksum and start building again beginning from step 3 (```COPY requirements.txt /usr/src/paperless/```). Checkout Dockers [best practices section](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache) for further details.

A side effect of this is, that only your local files are considered. E.g. Docker would not notice a newly uploaded package to PyPI which would be accepted by an cached requirements.txt. (you can use ```docker build --no-cache=true``` to force rebuild)

ENTRYPOINT or CMD can be placed anywhere in the Dockerfile, the last occurrence wins for each of them. Since it takes around 2 seconds on my machine to process each of these instructions, I've put them a bit up, so we can profit from another speedup there.

PR #380, opened 2018-07-17
WIP : Document sharing

CkuT:
Hey,

I'm working on a functionality that allows to create an anonymous fetching link for a document.
Every time I want to send someone a scanned file, I have to go to the interface, download the document, write my email and add the attached file, which I found a bit complicated.   

This is how this basic version works :
1: Generate a SharedDocument object that references a document
2: Give the generated link to somebody
3: Once the link has been consumed (= document downloaded), the SharedDocument object is destroyed, preventing further downloads on this link.

I am very much opened to improvements, especially if you want some enhancements to fit with your needs ;)

Some kind of ideas I was thinking about :

- Allow multiple fetches on a link by adding a counter field in the SharedDocument model, and decrement it each time the document is downloaded 
- Make an archive with multiple documents


danielquinn:
Ooh, this sounds a lot like another project I did called [Korra](https://github.com/danielquinn/korra).  Building on top of that might be a lot easier than tinkering with Paperless since this was never really meant to be a means of sharing etc.

CkuT:
Oh yes, Korra seems to be nice ! There are many interesting options, I really like the deletion after a fixed time.

If paperless is not meant to share documents, I agree with you on maybe building another application/software to handle the downloads. My idea is to keep the paperless-file-sharing to one-click, nothing more.
I think it would be a shame to redevelop a lot of code and adapt it on the top of the paperless database, particularly if the paperless app already has the knowledge of the fields and how handling the files (encryption/decryption)

Do you think something like a "plugin", which will be enabled/disabled as the user wishes would be a better compromise ?

danielquinn:
The thing is, Paperless' primary goal is to make it easy for people to index and search their documents.  For [security reasons](https://fsfe.org/contribute/promopics/thereisnocloud-bluecolor-preview.png), it was never meant to be placed on a public server, let alone serve as a distribution hub of those documents to others.  You could of course, bolt-on a sharing function, but the limited overlap between the two functions would make that sort of thing a poor candidate to merge into core.

Paperless does however have some semblance of a plugin infrastructure that might help you out here.  For starters, I'm not seeing that you actually need any of the database features for this process: the name, correspondents, tags, and content aren't necessary for file sharing.  Indeed, there's not much you need Paperless to do other than be the place where the file lives.

So with this in mind, I think the best bet for you is to make use of the [document consumption hook](http://paperless.readthedocs.io/en/latest/consumption.html#hooking-into-the-consumption-process) and have it execute a script that:

* Talks to your file sharing service, sending the path to the file
* Emails you (or some other means of notification) that a file is now ready to be shared

That way, your sharing service (whether you're using Korra or something else) has everything it needs, the files aren't duplicated, and you're free to do whatever you want with your shared files.

The one exception is that if you want to automatically delete the file (and db entry) from Paperless.  For that you'd probably want to just use the REST API.  I think we have a `delete` option, but if we don't, we can add it.

Similarly, there's no need to hook into the document consumption process if you build your sharing service right on top of the Paperless REST API.  Your sharing app could have a flow that looks a little like this:

1. User logs in
2. They look at the list of available files to share (provided in the background via Paperless' REST API)
3. They select which file(s) they want to share.  The API provides the path and all the metadata you want.
4. Your app sends the email, provides the external link etc etc.

Any of this sounding good?

CkuT:
You make a good point on privacy and security aspects, and I understand you don't really want to merge it.
Unfortunately, I don't have much time at the moment to integrate an external service. For the moment, I'll close this PR, and I'll look around for some basic integration to an external app as soon as possible.

danielquinn:
Sorry to disappoint!  I appreciate the enthusiasm though :-)

PR #383, opened 2018-07-19
Stop tests from writing to the source tree

erikarvstedt:
While packaging paperless for [nixpkgs](https://github.com/NixOS/nixpkgs) I discovered that the tests write to the project source tree, which makes my package build fail.
Although I can work around this in the packaging process, I think it's best when tests treat the project source as immutable.

The offending tests are in `test_date.py` where a source dir is [used as a scratch dir](https://github.com/danielquinn/paperless/blob/master/src/paperless_tesseract/tests/test_date.py#L36) for `RasterisedDocumentParser`.
Other tests in the same file use a [scratch dir in /tmp](https://github.com/danielquinn/paperless/blob/master/src/paperless_tesseract/tests/test_date.py#L16) that gets removed via `tearDown` after each test.
So on first look, it may seem the source scratch dir has the purpose of retaining parser temp files for further inspection after a failed test.

But actually, none of the offending tests produce any temp files, only empty parser temp dirs.
(These empty dirs in `src/paperless_tesseract/tests/samples` don't show up in git's staging view and thus go unnoticed. You probably have hundreds of them.)
So we can simply change these tests to use /tmp. That's what this PR does.

### DIFF
--- a/src/paperless_tesseract/tests/test_date.py
+++ b/src/paperless_tesseract/tests/test_date.py
@@ -33,7 +33,7 @@ class TestDate(TestCase):
 
     @mock.patch(
         "paperless_tesseract.parsers.RasterisedDocumentParser.SCRATCH",
-        SAMPLE_FILES
+        SCRATCH
     )
     def test_date_format_2(self):
         input_file = os.path.join(self.SAMPLE_FILES, "")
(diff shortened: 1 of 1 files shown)


erikarvstedt:
Any objections to this PR?

danielquinn:
Sorry Erik, I've been out of country and dealing with my in-laws for about a month now.  I'd forgotten about this.  It looks fine to me though, so I'll merge it now.  Apologies for the delay.

erikarvstedt:
That's just fine, welcome back to hacking! :hugs:

PR #387, opened 2018-08-16
Enable CORS for localhost

LukaszSolo:
Hi,

I'm working on the new (minimalistic for now) frontend. As I don't want to change anything in the existing Django app (for now at least), I'm running it as a separate app that consumes the Django API.

To make it a replacement for current UI, I needed to do some tiny changes:

- enable CORS for localhost calls
- add Filter to allow API calls that can select Documents without any tag

I don't foresee any issues with the above, but if you think it's not a good idea, let me know!

### DIFF
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -61,6 +61,7 @@ INSTALLED_APPS = [
     "django.contrib.messages",
     "django.contrib.staticfiles",
 
+    "corsheaders",
     "django_extensions",
 
     "documents.apps.DocumentsConfig",
@@ -84,6 +85,7 @@ if os.getenv("PAPERLESS_INSTALLED_APPS"):
 MIDDLEWARE_CLASSES = [
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'corsheaders.middleware.CorsMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
@@ -91,6 +93,9 @@ MIDDLEWARE_CLASSES = [
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 ]
+
+# We allow CORS from localhosts
+CORS_ORIGIN_REGEX_WHITELIST = (r'^(https?:\/\/)?localhost(:[0-9]{4})?$', )
 
 # If auth is disabled, we just use our "bypass" authentication middleware
 if bool(os.getenv("PAPERLESS_DISABLE_LOGIN", "false").lower() in ("yes", "y", "1", "t", "true")):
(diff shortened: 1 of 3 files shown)


danielquinn:
This looks pretty good, and I'm stoked about a new frontend!  Travis has pointed out that you need to use `==` instead of `=`, so that needs to change before I can merge it for sure.

I have one other question though.  As some of our users host Paperless on a domain other than localhost and/or a different port, can this be made a little more accommodating?  I'm thinking maybe use what you've got as the default, and allow for using a different value by setting `PAPERLESS_CORS_WHITELIST`?

```python
 CORS_ORIGIN_REGEX_WHITELIST = (r'^(https?:\/\/)?localhost(:[0-9]{4})?$', ) 
```

LukaszSolo:
Hey, thanks for a very quick response!

I'm a little stumped about the `==` as the settings file should use `=` . I think I know the issue and I'll double check in a couple of hours.

I see your point about the `PAPERLESS_CORS_WHITELIST` I think it's a good idea, let me add that too.

Thanks!

danielquinn:
requirements.txt uses a format where you have to set one of `==`, `>=`, or something elaborate like `<x.y,>=a.b`.  However, now that I think about it, in this project requirements.txt file is actually generated from the `Pipfile` anyway.  So if you're familiar with pipenv (the new hotness in Pythonland) jus `django-cors-headers = "<2.5,>=2.4"` to `Pipfile` and then run `pipenv lock && pipenv lock -r > requirements.txt`.

Barring that, I can just do that right on master and accept the rest of your PR once we work out the `WHITELIST` question.

LukaszSolo:
Hi,

I fixed the requirements by adding missing `=`. I'm not too familiar with pipenv (at least for now ;) ) so I think it might be better to leave it to you if you have a moment.

I also removed the regex and replaced it with less-powerful, but harder to mess up list. It should cover most of the common use cases like a subdomain, different ports etc.

src/paperless/settings.py
```diff
@@ -92,6 +94,13 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 ]
 
+# We allow CORS from localhost:8080
+CORS_ORIGIN_WHITELIST = ("localhost:8080")
+_allowed_cors_hosts = os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS")
+if _allowed_cors_hosts:
+    CORS_ORIGIN_WHITELIST = tuple(_allowed_cors_hosts.split(","))
```
sbrunner: All the 4 lines are the same as... :
`CORS_ORIGIN_WHITELIST = os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "localhost:8080").split(",")`

LukaszSolo: I followed the code pattern of ALLOWED_HOSTS as I didn't want to introduce a different style of writing.  WDYT?

danielquinn: I like @sbrunner's suggestion, as it's more conscise.  I see what you mean about the ALLOWED_HOSTS though.  We can fix that later.  For now, having this in one statement is better, so let's go with that.

LukaszSolo:
Thank you! I'll let you know when I think the Vue frontend is in a state worth sharing :)

PR #391, opened 2018-08-26
Better interface when we have many tags

sbrunner:
=>
tags

### DIFF
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -137,6 +137,7 @@ class DocumentAdmin(CommonAdmin):
                     "tags_")
     list_filter = ("tags", "correspondent", FinancialYearFilter,
                    MonthListFilter)
+    filter_horizontal = ("tags",)
 
     ordering = ["-created", "correspondent"]
 


boringusername99:
This looks like a great add! I'm surprised no one else has mentioned sorting of tags to be alphabetical, both on this selection UI and in the document sidebar. Did you consider this enhancement at all?

sbrunner:
+1 to sort tags alphabetical, I should fond how to do that and purpose a new pull request :-)

danielquinn:
This is an excellent addition, thank you!

As per @boringusername99's suggestion, I've also added a default ordering for tags.

ddddavidmartin:
I might be a bit slow, how do you get to this new tags menu? Or is it only available after a certain number of tags? Thanks!

danielquinn:
The new menu should appear as soon as you restart the `runserver` regardless of the number of tags you've got.

ddddavidmartin:
Ah yes, thanks! It shows when changing the properties of a document. I thought this would change how documents can be filtered with tags.

PR #395, opened 2018-09-04
Add text file consumer

### LINKED ISSUE #223: Use paperless in a more general purpose (handle text files)
Hi @danielquinn , I see paperless has evolved since last time I checked, nice work :)

I'm currently searching for a way to store or index documents for a project, and paperless is surely the project that is the closest to what I have in mind.

However, and if I understand well, it only handles PDF or image files. Would it be interesting, for you and other users, to add support for text files (just copy the content in DB for search feature)? Paperless would need to recognize text files as such.

Also, maybe we could have an option to also add any type of file, even if no text can be extracted for search-feature (we would still have correspondents, title, date and tags).

jat255:
This PR adds the ability for Paperless to handle .txt, .md (as plain text), and .csv files (as that's what I had in my library). Similar to the PDF handling, it generates a thumbnail from the file, so it displays like any other. 

I haven't written formal tests for it, but it is working correctly in my local testing, with respect to consumption and serving the resulting files. 

As this is my first PR for paperless, please let me know if there's anything else needed.

Resolves #223


jat255:
I suppose this would resolve #223 as well

src/documents/models.py
```diff
@@ -365,51 +369,52 @@ class FileInfo:
         )
     )
 
+    formats = "pdf|jpe?g|png|gif|tiff?|te?xt|md|csv"
```
ddddavidmartin: Yes! That's better than having them duplicated all the way through.

src/paperless_text/parsers.py
```diff
@@ -0,0 +1,146 @@
...
+        # - XX-YY-ZZZZ with XX + YY being 1 or 2 and ZZZZ being 2 or 4 digits
+        # - XX. MONTH ZZZZ with XX being 1 or 2 and ZZZZ being 2 or 4 digits
+        # - MONTH ZZZZ, with ZZZZ being 4 digits
+        # - MONTH XX, ZZZZ with XX being 1 or 2 and ZZZZ being 4 digits
+        pattern = re.compile(
+            r'\b([0-9]{1,2})[\.\/-]([0-9]{1,2})[\.\/-]([0-9]{4}|[0-9]{2})\b|' +
+            r'\b([0-9]{1,2}[\. ]+[^ ]{3,9} ([0-9]{4}|[0-9]{2}))\b|' +
+            r'\b([^\W\d_]{3,9} [0-9]{1,2}, ([0-9]{4}))\b|' +
+            r'\b([^\W\d_]{3,9} [0-9]{4})\b')
+
```
ddddavidmartin: Is this date matching copied from somewhere else in the code base? Would it be worth pulling it out and not duplicating it? It is quite a chunk and I reckon it would be good to have only one implementation of it to avoid issues in the future. Such as updating one and forgetting about the other one.

jat255: Ok, I moved the regex pattern into `documents/parsers.py`, and then have it imported into `paperless_tesseract` and `paperless_text`, so it will only have to be edited in one spot.

danielquinn:
Wow!  This is a lot of work @jat255!  Thanks so much for your contribution!  I've merged it into master and made one minor change: I renamed `pattern` to `DATE_REGEX` as it's now a constant rather than a local variable.

You probably don't know this, but @philippeowagner has been wanting this feature for a Very Long Time.  Sorry for the delay Philippe, but it looks like Joshua has you covered :-)

philippeowagner:
Awesome! Thanks @jat255 for creating this PR! 
It seems, this year Xmas comes a little earlier.

PR #396, opened 2018-09-04
Fix document checks with PostgreSQL and MySQL backends.

dubit0:
When running PostgreSQL or MariaDB/MySQL backends, almost all django management commands fail as long as the database is not yet populated. 

The reason seems to be that for those backends a ProgrammingError is raised in case a table does not exist. This error is, however, not properly caught in documents/checks.py. 

This patch fixes the error handling in the check.


danielquinn:
Hey thanks for this!  It's not something I would have thought to check for.

PR #399, opened 2018-09-05
Speed up thumbnail generation for PDFs

jat255:
Since we only use the first page of the PDF for thumbnail generation, the current method of converting the entire pdf to a series of .png images is wasteful, and can bog down the consumer if dealing with a very long (page-wise) PDF.

This PR uses the `convert` bracket syntax to load only the first page, and get the thumbnail from that instead of doing all the pages.


danielquinn:
This is a simple, and super-handy improvement!  Thanks!

PR #401, opened 2018-09-06
add migrate commande to docker update process

ahyear:


danielquinn:
Nice catch!  Thanks :-)

PR #404, opened 2018-09-12
ADD Paperless_inline_doc variable to docker

ahyear:


danielquinn:
I think this PR points out a glaring problem in how we're currently handling configuration variables in Paperless between the "bare metal" install and Docker-based one.

It's not entirely obvious, but everything defined in `paperless.conf` can be defined in `docker-compose.env`, and I'm not keen on duplicating variables & documentation from `paperless.conf.example` to `docker-compose.env.example`.

So instead, what I've done now is cleaned out all duplicate values from `docker-compose.env.example` and replaced them all with an explanation that this file can contain all the stuff defined in the other config.  It's not very clean, but given my limited knowledge of docker-compose, it's all I've got.

Ideally what I'd like to do is to have docker-compose setup to read docker-related environment stuff (`USERMAP_UID`, `PAPERLESS_OCR_LANGUAGES`, etc) from `docker-compose.env`, and then read everything else from `paperless.conf`, but since I don't know how to do that, this change will have to do.

So, TL;DR: I appreciate what you were trying to do here, but I'm trying to keep duplication to a minimum, so instead I've tried to better document what's going on.

Issue #406, opened 2018-09-16
consuming file fails due to several reasons


daedadu:
I'm german and trying to scan a document. So you bet there are some nasty characters in this file like "ä,ü" and so on ;). I'm running this in docker, this is my first docker experience.

When running `./manage.py document_consumer` i get the following:

```
bash-4.4# ./manage.py document_consumer
Starting document consumer at /consume with inotify
Parsers available: RasterisedDocumentParser
Consuming /consume/Scannen.pdf
Processing sheet #1: /tmp/paperless/paperless-pjkrcr4l/convert-0000.pnm -> /tmp/paperless/paperless-pjkrcr4l/convert-0000.unpaper.pnm
[pgm_pipe @ 0x5650674ba700] Stream #0: not enough frames to estimate rate; consider increasing probesize
out of deviation range - NO ROTATING
[image2 @ 0x5650674bb720] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
[image2 @ 0x5650674bb720] Encoder did not produce proper pts, making some up.
OCRing the document
Parsing for eng
Parsing for deu
Processing sheet #1: /tmp/paperless/paperless-pjkrcr4l/convert-0000.unpaper.pnm -> /tmp/paperless/paperless-pjkrcr4l/convert-0000.unpaper.unpaper.pnm
unpaper: error: output file '/tmp/paperless/paperless-pjkrcr4l/convert-0000.unpaper.pnm' already present.

Try 'man unpaper' for more information.
[pgm_pipe @ 0x5628ef2f3700] Stream #0: not enough frames to estimate rate; consider increasing probesize
out of deviation range - NO ROTATING
[image2 @ 0x5628ef2f4720] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
[image2 @ 0x5628ef2f4720] Encoder did not produce proper pts, making some up.
PARSE FAILURE for /consume/Scannen.pdf: Unpaper failed at ('unpaper', '/tmp/paperless/paperless-pjkrcr4l/convert-0000.pnm', '/tmp/paperless/paperless-pjkrcr4l/convert-0000.unpaper.pnm')
```

Replacing `command_args = unpaper, pnm, pnm.replace(".pnm",".unpaper.pnm")` in `parsers.py`

by

`command_args = unpaper,"--overwrite", pnm, pnm.replace(".pnm",".unpaper.pnm")`

gets me another step further.

```
[...]

PR #413, opened 2018-09-24
Fix issue where tesseract langages weren't installed properly

euri10:
Fixes https://github.com/danielquinn/paperless/issues/406


danielquinn:
This is not something I would have figured out easily, thank you!

PR #419, opened 2018-10-08
Let unpaper overwrite temporary files.

ddddavidmartin:
I'm not sure what the circumstances are, but it looks like unpaper can attempt to write a temporary file that already exists [0]. This then fails the consumption. As per daedadu's comment simply letting unpaper overwrite files fixes this.

[0]
unpaper: error: output file '/tmp/paperless/paperless-pjkrcr4l/convert-0000.unpaper.pnm' already present.
See https://github.com/danielquinn/paperless/issues/406#issue-360651630

### DIFF
--- a/src/paperless_tesseract/parsers.py
+++ b/src/paperless_tesseract/parsers.py
@@ -218,7 +218,8 @@ def run_convert(*args):
 
 def run_unpaper(args):
     unpaper, pnm = args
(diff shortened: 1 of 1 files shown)


ddddavidmartin:
I had the same issue on my setup, and this fixes it.

danielquinn:
Works for me!  Thanks for staying on top of this.

Issue #416, opened 2018-10-03
AttributeError at /admin/documents/document/


goose-ws:
When I pull up my /AttributeError at /admin/documents/document/ I get the following error message:

`Request Method: GET  

http://192.168.1.100:7412/admin/documents/document/  

1.11.13  

AttributeError  

'DocumentAdmin' object has no attribute 'get_changelist_instance'  

/home/goose/bin/paperless/src/documents/admin.py in changelist_view, line 212  
/usr/bin/python3  

3.5.3  

['/home/goose/bin/paperless/src',  '/usr/lib/python35.zip',  '/usr/lib/python3.5',  '/usr/lib/python3.5/plat-x86_64-linux-gnu',  '/usr/lib/python3.5/lib-dynload',  '/home/goose/.local/lib/python3.5/site-packages',  '/usr/local/lib/python3.5/dist-packages',  '/usr/lib/python3/dist-packages']  

Wed, 3 Oct 2018 15:57:10 +0000`

![image](https://user-images.githubusercontent.com/874014/46423259-305d0800-c704-11e8-80bb-c7f6151376ff.png)

Here is the traceback it provides:
http://dpaste.com/3ABY9YA.txt

I checked out issue #414 and #415 -- However, I don't believe it's a dates issue, as I originally had no documents at all and was getting the error. After processing in 3 documents, I see the dates as:

`goose@eddard:~/bin/paperless/src$ echo "SELECT created, modified, added FROM  documents_document ORDER BY added" | ./manage.py dbshell  

2018-01-08 00:00:00|2018-10-03 15:56:34.695460|2018-10-03 15:56:34.695273  

2017-10-01 00:00:00|2018-10-03 15:56:38.391423|2018-10-03 15:56:38.391252  

2018-03-27 20:55:34.355462|2018-10-03 15:56:38.896063|2018-10-03 15:56:38.895864  

goose@eddard:~/bin/paperless/src$`

Would thoroughly appreciate any help :)

---

Edit: Please disregard the 30 revisions on this. I can't get line breaks to work inside the code blocks, sorry.

Forceu:
I have the same problem, the traceback is a followed:

```
Environment:


Request Method: GET
Request URL: http://banana:8000/admin/documents/document/

Django Version: 1.11.13
Python Version: 3.5.2
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
[...]

PR #420, opened 2018-10-08
Hotfix for issue #416

Forceu:
This comments out a part that was added in the last update and causes the webserver to stop serving files. It is just a temporary fix so that users that are affected by this issue can continue using paperless until the cause is fixed.

Issue #416


Forceu:
Issue fixed after updating requirements

PR #421, opened 2018-10-08
Clarify forgiving ocr handling

ddddavidmartin:
Good day,

I have had a document fail to consume because no language could be detected. I had to look it up in the code as it wasn't clear to me what to do about it. With this pull request, we

- mention the `PAPERLESS_FORGIVING_OCR` option in the error message as a hint on how to continue with document consuming even in case of language detection errors
- add the `PAPERLESS_FORGIVING_OCR` option to the example config so that it is documented

Thanks!


danielquinn:
This is excellent, thank you!

Issue #348, opened 2018-04-28
ubuntu 14.04


ovv:
While working on #342 I wanted to drop the upstart documentation (14.04 is quite old and less documentation is easier to keep up to date) but I noticed the Vagrantfile is still based on 14.04.

I'm not familiar with Vagrant but I guess it would be beneficial to update it to a more recent ubuntu release

danielquinn:
One of the biggest problems with this project is that I don't really have a means of knowing what people are using, and *how* they're using it, so it's tough to know if dropping Vagrant support is going to bother anyone.

Frankly, I think that in 2018, Vagrant has more-or-less given way to Docker in terms of developer preference, so how 'bout this: drop the Vagrant documentation completely as part of #342, and when that's merged, I can toss out the Vagrantfile as well.  If there's anyone still using Vagrant, they'll notice, complain, and then I can see how many people are really using it.  If it's a lot, maybe I can work with those users to have one of them maintain a Vagrantfile for this project rather than keeping it part of core.

PR #425, opened 2018-10-13
Remove Vagrant docs

mrwacky42:
* Vagrant does not seem to have any libvirt boxes for Ubuntu any more.
* Vagrant 2 was released a year ago, but vagrant-libvirt only claims
  to support up to Vagrant 1.8.

"Fixes" #348?


danielquinn:
One less thing to support!  Thanks for this :-)

PR #432, opened 2018-10-31
Added missing ; to nginx config

deanpcmad:
Without these nginx fails to start


danielquinn:
Thanks for catching this!  Merging now :-)

PR #441, opened 2018-11-16
Update gunicorn commands

jat255:
I had previously been using Django's built-in server, but I just switched to nginx+gunicorn (much faster!)

On my system, the gunicorn command (and the service files) needed some modification to run because of a `ModuleNotFoundError`. The solution was found [here](https://stackoverflow.com/questions/47710428/gunicorn-no-module-named-path-to-my-django-project). I believe all the relevant files have been updated, if anyone wants to test them out.

### DIFF
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -366,7 +366,7 @@ The gunicorn server can be started with the command:
 
 .. code-block:: shell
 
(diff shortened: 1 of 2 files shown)


danielquinn:
Honestly, when it comes to Gunicorn configs, I'm really not sure any more.  If this works for you, that's good enough for me :-)

PR #442, opened 2018-11-17
Add example javascript override for tag colour display

jat255:
This PR does two things:

1) Creates an `overrides` folder in the project root with a README where example overrides for `.css` and `.js` files can be stored (and so provide a location to which people can send PRs)
2) Adds an override that uses jQuery to display a little preview of a tag's color when adding a new tag (since I could not figure out how to do this within the Django interface itself)

Here's how it looks in action after adding the code to `overrides.js`:

demo

It does not yet do anything on the "edit tag" page, but I'm looking into that.

### DIFF
--- a/src/documents/static/js/colours.js
+++ b/src/documents/static/js/colours.js
@@ -1,7 +1,6 @@
 // The following jQuery snippet will add a small square next to the selection
 // drop-down on the `Add tag` page that will update to show the selected tag
-// color as the drop-down value is changed. Copy and paste it into
-// ``PAPERLESS_MEDIADIR/overrides.js`` to see the effects.
+// color as the drop-down value is changed.
 
 if (django.jQuery("#id_colour").length) {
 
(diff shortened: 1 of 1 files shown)


jat255:
Ok, support for the edit tags page has been added too. Here's what it looks like:

peek 2018-11-17 21-35

jat255:
Also, perhaps we should put a link to the new overrides folder in the documentation on the [customizing](https://paperless.readthedocs.io/en/latest/customising.html) page?

overrides/README.md
```diff
@@ -0,0 +1,11 @@
+# Customizing Paperless
```
danielquinn: This is a great idea, and something I really need to remember to more of: putting documentation *in* the file structure to give people "signposts" rather than just assuming they've read all the documentation.  Thanks for this.

overrides/tag_colour_preview.js
```diff
@@ -0,0 +1,67 @@
+// The following jQuery snippet will add a small square next to the selection
```
danielquinn: I **love** what this does, but I'm not sure that this is the best place for it

This is the sort of thing that probably shouldn't live in `overrides/` but should rather just be part of core.  I'm happy to move this to `documents/static/js/colours.js` or something so that it's the default for everyone.  I can't see any reason why we wouldn't want this functionality by default.  If you're cool with that, please modify this PR to place this in `documents/static/js/colours.js` (you'll need to create the `js` directory) and I'll edit the Python code to include the `.js` file directly, bypassing the need for an override.

jat255: Ok, done.

jat255:
I'll have to depend on you to add the code to load the `colours.js` file, since I'm not sure where in the source is the best spot for that

PR #450, opened 2018-12-02
Fix date test sample image

erikarvstedt:
##### Commit message
The previous version of `tests_date_3.png` had too much spacing
between the `0` and the `8` glyphs, which resulted in the year getting
parsed as `200 8` in Tesseract 3.05.00 (+ tessdata 3.04.00).
This caused the date parsing test to fail.

@danielquinn, it's weird that this test worked for you and failed for me (in both 'english' and 'german' ocr runs), but the excessive spacing is [evident](https://github.com/danielquinn/paperless/compare/master...erikarvstedt:fix-parser-test?expand=1&short_path=093e9d7#diff-093e9d7cac8f2a941f4e4a4f654094d5).

By the way, Travis seems to have stopped working. The latest [test](https://travis-ci.org/danielquinn/paperless/branches) of the master branch was 9 months ago. Or am I missing something?

Edit: Added a fix for Travis so that this PR is properly tested. Master branch testing is still broken.
Edit 2: Whoah, my Travis fix triggered a load of other [test failures](https://travis-ci.org/danielquinn/paperless/jobs/462453066). Why is 'Portuguese' suddenly detected as a doc language? 
There's something very weird going on with the dependencies on Ubuntu/Travis.
Should I fix Travis to use the Paperless Docker environment for testing?


danielquinn:
I think the root of the problem is in the use of Lorem Ipsum everywhere, which is making the language detection flip out.  What's worse, is that the unpredictable nature of how Tesseract might parse a file pixel-by-pixel means that a prediction can't even be counted on to be consistent.  I've seen Portugese, Catalan, etc.

The only fix I've seen to work is to re-do all sample files to use English for the English tests and German for the german ones.  Even then, it's probably a good idea to use `override_settings` for the default language so that things don't get accidentally confused.  As the tests are for date detection, I don't think controlling for language in this case is a bad idea anyway.

I don't have a lot of time lately, but if you wanna re-roll the images in the right languages, I'll happily merge it.

As for this PR, it's sort of a band-aid, as it'll account for cases where language guessing works out correctly to get German.  As that's not a guarantee given the current state of things, there's no way to be sure this will always fix things, but it's better than nothing.

Your suggestion however of getting Travis to use the Docker container is an excellent one.  I don't know how to do that, as my Travis-foo is limited, but if you've got the skillz, I'm happy to merge that too.

PR #451, opened 2018-12-02
Add DBHOST & DBPORT parameters to settings

### LINKED ISSUE #445: Is it possible to use a Postgis db with Paperless in Docker
Here
https://github.com/danielquinn/paperless/blob/d825667c9bffe615b753451c5ddbe50d693bc978/src/paperless/settings.py#L147-L154
I see that we can configure the username and the password, but we can't change the host shouldn't we add optional `PAPERLESS_DBHOST` and `PAPERLESS_DBPORT`?

speshak:
This resolves #445.  This PR also includes updates requirements.txt & Dockerfile required to actually use Postgres from the docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,10 @@ ENV PAPERLESS_EXPORT_DIR=/export \
     PAPERLESS_CONSUMPTION_DIR=/consume
 
 
-RUN apk update --no-cache && apk add python3 gnupg libmagic bash shadow curl \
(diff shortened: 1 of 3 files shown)


sbrunner:
For me the Travis errors are not related on those changes, I get the same errors on master ...

danielquinn:
Yeah the tests are proving to be a little problematic/finicky lately.  This looks good though.  Thanks for the contribution!

sbrunner:
Thanks for all :-)

PR #468, opened 2018-12-31
Reference PAPERLESS_OCR_ALWAYS in example config file.

ddddavidmartin:
This setting was introduced when support for retrieving the text layer
from documents was added. Having it in the example config makes it more
clear that it exists.


danielquinn:
Ooh, good call.  Thanks for noticing and taking the initiative on this.

PR #471, opened 2019-01-03
Exposing documents' "added" timestamp via Rest API.

khrise:
I recognized that the Rest API doesn't expose the "added" timestamp of documents. This PR fixes that.
I hope it's ok to create a PR without opening an issue before.


danielquinn:
Hey thanks for the PR!  Yeah it's totally fine to open a PR without a related issue, so don't worry about it :-)

PR #475, opened 2019-01-12
Catch IMAP connection errors

### LINKED ISSUE #474: Consumer stops working
Hello,
After several times the consumer stops working with the following error.

`aperless-consumer.service - Paperless consumer
   Loaded: loaded (/home/paperless/script/paperless-consumer.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Fri 2019-01-11 12:58:45 CET; 2h 43min ago
  Process: 1233 ExecStart=/usr/bin/python3 /home/paperless/paperless/src/manage.py document_consumer (code=exited, status=1/FAILURE)
 Main PID: 1233 (code=exited, status=1/FAILURE)

Jan 11 12:58:44 nuc python3[1233]:     sock = IMAP4._create_socket(self)
Jan 11 12:58:44 nuc python3[1233]:   File "/usr/lib/python3.6/imaplib.py", line 289, in _create_socket
Jan 11 12:58:44 nuc python3[1233]:     return socket.create_connection((host, self.port))
Jan 11 12:58:44 nuc python3[1233]:   File "/usr/lib/python3.6/socket.py", line 704, in create_connection
Jan 11 12:58:44 nuc python3[1233]:     for res in getaddrinfo(host, port, 0, SOCK_STREAM):
Jan 11 12:58:44 nuc python3[1233]:   File "/usr/lib/python3.6/socket.py", line 745, in getaddrinfo
Jan 11 12:58:44 nuc python3[1233]:     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
Jan 11 12:58:44 nuc python3[1233]: socket.gaierror: [Errno -2] Name or service not known
Jan 11 12:58:45 nuc systemd[1]: paperless-consumer.service: Main process exited, code=exited, status=1/FAILURE
Jan 11 12:58:45 nuc systemd[1]: paperless-consumer.service: Failed with result 'exit-code'.`

Any idea to fix this?

syntonym:
imaplib.IMAP4_SSL raises a socket.gaierror when something goes wrong with connecting
to PAPERLESS_CONSUME_MAIL_HOST. Later MailFetcherErrors are excepted, but socket.gaierror
are not, so instead raise a MailFetcherError. Because socket.gaierrors can have numerous
reasons (gaierror wraps EAI_* errors from getaddrinfo, so see man getaddrinfo EAI_*),
include the error description in the exception.

Fixes #474

### DIFF
--- a/src/documents/mail.py
+++ b/src/documents/mail.py
@@ -216,7 +216,11 @@ class MailFetcher(Loggable):
         return r
 
     def _connect(self):
-        self._connection = imaplib.IMAP4_SSL(self._host, self._port)
+        try:
+            self._connection = imaplib.IMAP4_SSL(self._host, self._port)
+        except OSError as e:
+            msg = "Problem connecting to {}: {}".format(self._host, e.strerror)
+            raise MailFetcherError(msg)
 
     def _login(self):
 


ovv:
This looks good :+1: 

Maybe we can catch all socket.* errors (`socket.herror`, `socket.gaierror`, `socket.timeout`).
See: https://docs.python.org/3/library/socket.html#exceptions

ovv:
I'm not sure what python version paperless aims to support. If >= 3.3 then `OSError` sounds like the best solution to me.

syntonym:
I can't find any sources right now but I thought that virtually all linux distributions are on python3 >= 3.3. [As far as I see](https://devguide.python.org/devcycle/#end-of-life-branches) python 3.0-3.3 is end of life, so assuming python >= 3.3 seems unproblematic to me.

MasterofJOKers:
Additionally, `django = "<2.1,>=2.0"` is a dependency. Django 2.0 supports 3.4, 3.5, 3.6, 3.7 ([see FAQ](https://docs.djangoproject.com/en/2.0/faq/install/#faq-python-version-support)). So even python >= 3.4 can be assumed.

syntonym:
Should this include a test? I'm not super sure how to test that, besides spinning up a IMAP server and then manipulating the connection, which seems "overkill".

syntonym:
To me that seems to stray into "adding a test for adding a test" territory, but if you you it would be worth it I can add it.

MasterofJOKers:
My point was just, that writing a test would be possible without an imap server. I agree that for this exception handler, probably no test is necessary, if we don't aim for 100% test coverage.

danielquinn:
Honestly, I'd like to aim for 100% test coverage (or at least explicitly set `#param nocover` for code we don't want to worry about).  As a friend of mine has recently brought me over to the 100% camp says, it makes for a good psychological barrier: dipping from 100% to 99.99% forces you to consider why it's ok not to cover this particular case, while dipping from 99.99% to 99.98% doesn't seem so bad.

With all that said though, I'm a pragmatist when it comes to this project and if there's consensus here, that this PR should be merged, then I'll hit the Big Green Button.  @ovv, @MasterofJOKers, @syntonym, what say you?

ovv:
I agree with you @danielquinn , tests would be better but we should also make sure that it doesn't drive contribution away. IMO a good starting point is to require tests for new features and important bugfixes. In this case I don't think a test would add anything.

:shipit:

syntonym:
While test coverage is important I think even 100% of source lines covered by unit tests is "not enough". Adding a unit tests which throws a `socket.gaierror` is easy and improves test coverage, but a `socket.timeout` would still crash the consumer. This is even true with 100% test coverage (counted by lines covered by unit tests), so that even 100% test coverage doesn't guarantee anything. 

As this PR is of low impact I'd advocate for merging it while I'm looking into writing some wider unittests covering the consumption process as ovv suggested.

danielquinn:
Thanks for all the input guys.  I'll merge this now.

PR #476, opened 2019-01-12
Install psycopg2 in the Docker container

sbrunner:
Just run `pipenv install psycopg2`

I see that all the packages will be updated, is it acceptable?

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -34,6 +34,7 @@ pytest-django = "*"
 pytest-sugar = "*"
 pytest-env = "*"
 pytest-xdist = "*"
+psycopg2 = "*"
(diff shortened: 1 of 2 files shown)


sbrunner:
UP?

danielquinn:
This looks fine to me, but as this updates a bunch of other packages, I don't want to merge it straight-away.  I'm going to leave this open for a little bit in the hopes that a few others chime in with a "it works for me".

Our tests pass, but as our tests are currently few, I don't think they're enough to :+1: a PR like this.

sbrunner:
Or I can clean it you prefer :-)

PR #478, opened 2019-01-18
Update requirements.rst

wiwie:
added missing requirement optipng.

If this is not installed, running ./manage.py createsuperuser will emit the following warning:

> WARNINGS:
> ?: Paperless can't find optipng. Without it, consumption is impossible.
> 	HINT: Either it's not in your ${PATH} or it's not installed.


danielquinn:
Ooh, thanks for noticing!

Issue #350, opened 2018-05-04
Cannot set email comsumer when using Docker


HesselTjeerdsma:
Hi!

How do I correctly setup the imap consumer when using docker? Now, I tried to copy all email related settings from the paperless.conf.exampe to docker-compose.env but then I get the error:  
```

[Errno -8] Unrecognized service
--
```

Here is a copy of my docker-compose.env where I have removed all sensitive information:

```
# Environment variables to set for Paperless
# Commented out variables will be replaced by a default within Paperless.

# Passphrase Paperless uses to encrypt and decrypt your documents
PAPERLESS_PASSPHRASE=

# The amount of threads to use for text recognition
PAPERLESS_OCR_THREADS=4

# Additional languages to install for text recognition
PAPERLESS_OCR_LANGUAGES=nld

# You can change the default user and group id to a custom one
# USERMAP_UID=1000
# USERMAP_GID=1000

PAPERLESS_CONSUME_MAIL_HOST="imap.gmail.com"
PAPERLESS_CONSUME_MAIL_PORT="993"
PAPERLESS_CONSUME_MAIL_USER=""
PAPERLESS_CONSUME_MAIL_PASS=""
PAPERLESS_EMAIL_SECRET=""
```

erikarvstedt:
Is this a Docker or a Python error? In either case, please post the full backtrace/output.

danielquinn:
@ahyear just posted a full backtrace in #366 and witha little searching online, I have a guess.  Can one of you try changing these lines in `src/documents/mail.py` from this:

```python
        self._host = os.getenv("PAPERLESS_CONSUME_MAIL_HOST")
        self._port = os.getenv("PAPERLESS_CONSUME_MAIL_PORT")
        self._username = os.getenv("PAPERLESS_CONSUME_MAIL_USER")
        self._password = os.getenv("PAPERLESS_CONSUME_MAIL_PASS")
        self._inbox = os.getenv("PAPERLESS_CONSUME_MAIL_INBOX", "INBOX")
```

to this?

```python
        self._host = os.getenv("PAPERLESS_CONSUME_MAIL_HOST", "").strip()
        self._port = os.getenv("PAPERLESS_CONSUME_MAIL_PORT", "").strip()
        self._username = os.getenv("PAPERLESS_CONSUME_MAIL_USER", "").strip()
        self._password = os.getenv("PAPERLESS_CONSUME_MAIL_PASS", "").strip()
[...]

PR #480, opened 2019-01-20
closes #479

### LINKED ISSUE #479: paperless in lxc
Make paperless not only available in bare metal and docker, but also via Linux Containers. 
Add first install screenshots.
Add scanner recommendation of Brother MFC-J5910DW.

PR to follow.

bmsleight:
To close #479


danielquinn:
This all looks pretty good, and I especially like the work you've put in to include config file examples etc.  Travis is still barking about the sphinx build though -- it looks like there's a problem with one of the screenshot files.  Can you fix this?

ovv:
I'm not sure that we need a separate documentation for lxc, it should be pretty common with the `bare metal` one. Maybe we should rename to include `bare metal`, `virtual machine` and `lxc/lxd` ?

An installation script is a very good idea :+1:

syntonym:
Making installation easier is a good goal, but wouldn't a "proper" apt package achieve the same goal with more "standard" tools? Properly packaging would also add additional features like uninstalling without having to reinvent them in script form. Or are there steps in the install script which can not be archieved by a package (like adding users, generating the random key, copying stuff to `/etc/apache2/sites-available/`?

ovv:
An APT packages would be great but it's probably a lot more work.
> copying stuff to `/etc/apache2/sites-available`

I don't think an APT package would be allowed to do that since people might want to use nginx or any other webserver instead.

I believe having some install script would be a good step forward. Personally I am using an [ansible playbook](https://github.com/ovv/ansible-role-paperless) to manage my setup.

bmsleight:
Updated screenshots.rst 
Love the Travis builds
Python 3.4-3.6 passed - https://travis-ci.org/danielquinn/paperless/jobs/482554384 
Any idea why Python 3.7 failed ? https://travis-ci.org/danielquinn/paperless/jobs/482554387
ConnectionResetError: [Errno 104] Connection reset by peer ?

bmsleight:
@ovv  "I'm not sure that we need a separate documentation for lxc"
Agree :) - I wrote it as a third section: https://github.com/bmsleight/paperless/blob/master/docs/setup.rst#installation--configuration

@ovv I really need to try and learn ansible 

Did not expect so nice feedback on the install script, I could look at splitting it up in to sections and make optional for apache or nginx or the ftp users.

@syntonym apt-get would be lovely. The steps that make it hard is mixing of apt-get and pip. I have only done some very simple debian package (many, many years ago). So I would find it hard. (Hey but love to be user for it!) But I am not convinced it would get much use over the docker, ansible or lxc script. As I would expect most users would would not mix the paperless machine/container with much else. 

Thanks for the lovely feedback.
Brendan

ovv:
> Agree :) - I wrote it as a third section: https://github.com/bmsleight/paperless/blob/master/docs/setup.rst#installation--configuration

My point is that a third section might not be required. Any reason your script would not work on a bare metal or virtual machine ? The idea being to include your script in the bare metal instruction.

ovv:
> Python 3.4-3.6 passed - https://travis-ci.org/danielquinn/paperless/jobs/482554384
Any idea why Python 3.7 failed ? https://travis-ci.org/danielquinn/paperless/jobs/482554387
ConnectionResetError: [Errno 104] Connection reset by peer ?

From what I see after the restart 3.4 is failing. It looks like some issue in Travis. I'm not sure if you have the right to restart the build @bmsleight or only @danielquinn can do it ?

bmsleight:
> Any reason your script would not work on a bare metal or virtual machine
Ah! Yep should work. The bare metal instructions are good for experts, as my script is a bit cavalier with pip (no virtual-env) and apache2 default config. (But fixable if script in sections and optional parts)

I can restart by closing re-opening the pull request.  ....

danielquinn:
Sorry I'm late to this conversation, but my kid was *just* born last week so I've had my hands full.

@bmsleight, this is some great work, especially the documentation changes, which I always appreciate.

Your install script is pretty handy, but I don't think it makes sense to roll it out as "this is the way you do Paperless with lxc" as it's made some rather rigid assumptions like expecting the user to base their sutff on Ubuntu, and standardise on Apache & proftpd.  There's nothing *wrong* with these assumptions per se, just that having the documentation say "use this install script" really ties their the user's hands.

To @ovv's point about having a bare-metal install script, that's ambitious, but I think to do that, we'd have to see what other similar projects do to work with/around the multitude of configurations out there.  Maintaining a script that's smart enough to detect:

* The Linux distribution (and all it's quirks around file locations)
* The dependent software choices like webserver & file upload (smb/ftp/sftp etc)

I'm not sure that's even possible.

In the short term, I think the best option is to take the install script and move it from `scripts/lxc` into `docs/examples/lxc/` and then change the documentation to *suggest* something like that file.  This will let people set things up the way they like without saddling this project with more code to maintain so that it works with everyone's stuff.

@bmsleight, I appreciate your patience on this, but with such a big change, I want to be careful about what this project takes on.  What do you think?

bmsleight:
@danielquinn 
> kid was just born last week 

Feeble excuse :) Congratulations. Little ones can generate lots of paperwork.


> documentation say "use this install script" really ties their the user's hands.

Agree 


> move it from scripts/lxc into docs/examples/lxc/ and then change the documentation to suggest something like that file.

Will do 

>  What do you think?

I like the reasoning.  I will play with the files and update the PR

bmsleight:
Troubleshooting note can also close last comment on #350

erikarvstedt:
Could you break up your changes into logical commits, each describing a single change or topic you're introducing?
This helps reviewing now and at a later point (especially when it comes to code changes).

Here's a possible way of organizing the changes into commits:
- Add lcx installation method
- Docs: Add web interface screenshots
- Docs: Explain how to view consumer logs
- Docs: Add MFC-J5910DW to recommended scanners

bmsleight:
See #499 #500 #501

Issue #411, opened 2018-09-23
cache thumbnails in browser


Findus23:
related to #410 

Another way to massively speedup the page load (after the first time) would be to cache the thumbnails and only load them when they changed (via etag or something like this).

Now that I think about it, thumbnails should normally never change so running it behind behind nginx and adding a cache header for `/fetch/thumb/` should maybe also be enough.

MasterofJOKers:
+1 for this, because using paperless over a slow connection (when not at home via VPN) is a real pain with thumbnails are reloaded every time.

I've got some changes for this, but they're probably not clean enough for a pull-request, so I just leave them here as a hint what might be done:
```diff
diff --git a/src/documents/views.py b/src/documents/views.py
index e297e09..b4146be 100644
--- a/src/documents/views.py
+++ b/src/documents/views.py
@@ -1,6 +1,7 @@
 from django.http import HttpResponse, HttpResponseBadRequest
 from django.views.generic import DetailView, FormView, TemplateView
 from django_filters.rest_framework import DjangoFilterBackend
+from django.utils.cache import patch_response_headers
 from paperless.db import GnuPG
 from paperless.mixins import SessionOrBasicAuthMixin
 from paperless.views import StandardPagination
@@ -51,10 +52,12 @@ class FetchView(SessionOrBasicAuthMixin, DetailView):
         }
 
         if self.kwargs["kind"] == "thumb":
-            return HttpResponse(
+            response = HttpResponse(
                 self._get_raw_data(self.object.thumbnail_file),
                 content_type=content_types[Document.TYPE_PNG]
             )
+            patch_response_headers(response)
+            return response
 
         response = HttpResponse(
             self._get_raw_data(self.object.source_file),
diff --git a/src/paperless/settings.py b/src/paperless/settings.py
index 06cc180..833a316 100644
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -280,3 +280,5 @@ FY_END = os.getenv("PAPERLESS_FINANCIAL_YEAR_END")
 
 # Specify the default date order (for autodetected dates)
 DATE_ORDER = os.getenv("PAPERLESS_DATE_ORDER", "DMY")
+
+CACHE_MIDDLEWARE_SECONDS = 3600
```

[...]

PR #481, opened 2019-01-21
Add Cache-Control header for thumbnails

CkuT:
This drastically optimizes admin interface loading by telling the browser to cache thumbnails. The max-age recommendation is 1 year according to [rfc2616 ](https://www.ietf.org/rfc/rfc2616.txt)

### DIFF
--- a/src/documents/views.py
+++ b/src/documents/views.py
@@ -2,6 +2,7 @@ from django.http import HttpResponse, HttpResponseBadRequest
 from django.views.generic import DetailView, FormView, TemplateView
 from django_filters.rest_framework import DjangoFilterBackend
 from django.conf import settings
+from django.utils import cache
 
 from paperless.db import GnuPG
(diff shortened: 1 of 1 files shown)


MasterofJOKers:
Great PR. Would you mind adding a `Closes #411` to the commit message?
Additionally, we've discussed parts of this in #411 and I would be more comfortable with using django's  functionality like [`patch_response_headers()`]( https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.cache.patch_response_headers) or [`patch_cache_control()`]( https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.cache.patch_cache_control) method.

Personally, I also like having a small description (like your PR description) in the commit itself. Especially for your reference to the 1 year max-age recommendation.

CkuT:
Done !

danielquinn:
This is excellent!  Both the code change and the way you've both worked together to make it better.  Thank you to both!

PR #483, opened 2019-01-22
added fields to the correspondent and tag REST API

tsia:
i added those fields to expose them to the REST API so the API can show the full Data for Correspondents and Tags


danielquinn:
Looks like a simple request, and there's no additional db hits to get the info.  Merging now, thanks!

PR #492, opened 2019-01-26
Add and configure DjangoQL

sbrunner:
Use this plugin https://github.com/ivelum/djangoql#djangoql
to do advance search in the documents :-)


danielquinn:
As this PR was in conflict thanks to your other PR for psycopg2, I decided to take some time this morning and test out a proper update of all the dependencies with `pipenv update`.  The tests pass, the interface appears to be solid, so I'm giving all of this the :+1:.

As an aside, DjangoQL is amazeballs!  How did I not know about this before?

Issue #470, opened 2018-12-31
This project is bigger than me


danielquinn:
...and I don't have a whole lot of time to donate to it.

Since I started it back in 2015, Paperless has grown far beyond what I ever expected it to be.  With 5000+ stars on Github and a surprising popularity in Germany, issues are posted or commented on on an almost daily basis.

With the little time that I have (baby on the way!), I try to roll out fixes & features, but every change tends to add instabilities on platforms I don't personally use.  There are people using this project on bare-metal Linux installs, Docker installs on Synology boxes, cloud installations, Rasberry Pis, and FreeBSD machines -- and that's only the ones I know about from issues users have posted.

If this project is going to survive & grow, there has to be more people driving it.  Primarily, we need people who the rest of us can trust, but also people willing to do the boring work of adding tests and testing frameworks for the various platforms.

I'd like to transition Paperless to more of a collective effort, either as part of the @Jazzband, or some other means -- I'm open to suggestions.  I'm happy to stay on to audit pull requests or make suggestions, but it's becoming clear that Paperless needs more from me than I'm able to give it, and so I'm thinking we need to expand the pool.

Here's what I'm suggesting: I believe GitHub can be configured to require that merges be reviewed by one or more other collaborators before code is merged.  If I can figure out how to do this (maybe I need to create an organisation?  Advice on this is appreciated.) then we could have a system where we have more people who can create & merge more stuff, but where merges require review from at least one other person -- that'll keep people honest.

If GitHub can't do this, then I'm also happy to move to [GitLab](https://gitlab.com/) unless there are strong objections.

IMHO, the first mission of this collective should be to get us to 100% test coverage and to update Travis (or something else?) to execute those tests in different environments (if possible).  After that, I imagine some of you have your own priorities regarding where this should go.  I'm excited to see what you have in mind.

I'm tagging a few people specifically who have been more heavily involved with Paperless than others just to give them a heads-up.  If any of you lot are interested in taking a more primary role in this project, please comment below.  Maybe we can get an idea of what kinds of resources/interest is out there.  **This is not an exclusive list**, so if you don't appear here but feel you've contributed significantly (or just want to) it's because I didn't recognise your username from the autocomplete:

:+1: @Strubbl @philippeowagner @jat255 @ahyear @avichalp @BastianPoe @CkuT @ddddavidmartin @erikarvstedt @MasterofJOKers @pitkley @ovv

[...]

PR #494, opened 2019-01-29
fix parse error of some documents by using gs

### LINKED ISSUE #447: Parse error of some documents (not all)
I do get an error by comsuming some documents.
In the log file I got the following message:
```
PARSE FAILURE for /home/paperless/consume/20181201132749Z - Vodafone_0.pdf: Convert failed at ('convert', '-scale', '500x5000', '-alpha', 'remove', '/home/paperless/consume/20181201132749Z - Vodafone_0.pdf[0]', '/tmp/paperless/paperless-kljps7vb/convert.png')
Consuming /home/paperless/consume/20181201132749Z - Vodafone_0.pdf
Parsers available: RasterisedDocumentParser
```

-----------------

The status of the consumer process showes the following
```
Dez 01 13:39:19 nuc python3[31976]: Consuming /home/paperless/consume/20181201132749Z-Vodafone_0.pdf
Dez 01 13:39:19 nuc python3[31976]: convert-im6.q16: FailedToExecuteCommand `'gs' -sstdout=%stderr -dQUIET -dSAFER -dBATCH -dNOPAUSE -dNOPROMPT -dMaxBitmap=500000000 -dAlignToPixels=0 -dGridFitTT=2 '-sDEVICE=pngalpha' -dTextAlphaBits=4 -dGraphicsAlphaBits=4 '-r72x72' -dFirstPage=1 -dLastPage=1 
Dez 01 13:39:19 nuc python3[31976]: convert-im6.q16: no images defined `/tmp/paperless/paperless-k3id431m/convert.png' @ error/convert.c/ConvertImageCommand/3258.
Dez 01 13:39:19 nuc python3[31976]: PARSE FAILURE for /home/paperless/consume/20181201132749Z-Vodafone_0.pdf: Convert failed at ('convert', '-scale', '500x5000', '-alpha', 'remove', '/home/paperless/consume/20181201132749Z-Vodafone_0.pdf[0]', '/tmp/paperless/paperless-k3id431m/convert.png')
```

I already tried to reprint the pdf file into an other but it does not work.
Any proposals?

JensPfeifle:
Closes #447 when running paperless in docker


src/paperless_tesseract/parsers.py
```diff
@@ -45,13 +45,23 @@ def get_thumbnail(self):
...
+        gs_out_path = os.path.join(self.tempdir, "gs_out.png")
 
         # Run convert to get a decent thumbnail
+
+        # https://github.com/danielquinn/paperless/issues/447
+        # call gs first
+        environment = os.environ.copy()
+        cmd = ["gs", "-q", "-sDEVICE=pngalpha",
+               "-o", gs_out_path, self.document_path]
+        if not subprocess.Popen(cmd, env=environment).wait() == 0:
```
MasterofJOKers: Is there a reason you need to pass `env`? As I read the [documentation](https://docs.python.org/3/library/subprocess.html?highlight=subprocess#subprocess.Popen), it's the default behaviour to inherit the current process' environment.

JensPfeifle: I based this on the `Popen` call in `run_convert()`. But you are right, according to the docu it is probably unnecessary here. Would you prefer to have it without `env`?

MasterofJOKers: Yes, please. Any code that's not necessary takes concentration away from code that's relevant. You can also have a look at `run_unpaper()`, where `env` is not passed, as it's not needed.

src/paperless_tesseract/parsers.py
```diff
@@ -45,13 +45,22 @@ def get_thumbnail(self):
         """
 
         out_path = os.path.join(self.tempdir, "convert.png")
+        gs_out_path = os.path.join(self.tempdir, "gs_out.png")
 
         # Run convert to get a decent thumbnail
+
+        # https://github.com/danielquinn/paperless/issues/447
+        # call gs first
+        cmd = ["gs", "-q", "-sDEVICE=pngalpha",
```
pitkley: Rather than calling `gs` directly, you should add a new setting `GHOSTSCRIPT_BINARY` or `GS_BINARY`, as done for [`CONVERT_BINARY`](https://github.com/the-paperless-project/paperless/blob/6d8e4ad6bf92094149e2335a8f30cb976475fe70/src/paperless/settings.py#L261).

There are at least the following places where it should be added (there might be more I have missed):

* [`settings.py`](https://github.com/the-paperless-project/paperless/blob/6d8e4ad6bf92094149e2335a8f30cb976475fe70/src/paperless/settings.py#L261)
* [`paperless.conf.example`](https://github.com/the-paperless-project/paperless/blob/6d8e4ad6bf92094149e2335a8f30cb976475fe70/paperless.conf.example#L248)
* [`parsers.py`](https://github.com/the-paperless-project/paperless/blob/master/src/paperless_tesseract/parsers.py#L31) (only retrieval of the setting itself)

pitkley: Should we only use this as a fallback? I.e. should we try to execute ghostscript and fall back to the "old way" if it doesn't exist?

I was thinking about the function being something like this:

```python
    def get_thumbnail(self):
        """
        The thumbnail of a PDF is just a 500px wide image of the first page.
        """

        out_path = os.path.join(self.tempdir, "convert.png")
        gs_out_path = os.path.join(self.tempdir, "gs_out.png")

        # Extract the first PDF page as a PNG using Ghostscript, if available, to use as a thumbnail.
        # https://github.com/danielquinn/paperless/issues/447
        try:
            cmd = [
                self.GHOSTSCRIPT,
                "-q",
                "-sDEVICE=pngalpha",
                "-o", gs_out_path,
                self.document_path
            ]
            if not subprocess.Popen(cmd).wait() == 0:
                raise ParseError("Thumbnail (gs) failed at {}".format(cmd))
            thumbnail_source_path = gs_out_path
        except FileNotFoundError:
            thumbnail_source_path = "{}[0]".format(self.document_path)

        # Run convert to get a decent thumbnail
        run_convert(
            self.CONVERT,
            "-scale", "500x5000",
            "-alpha", "remove",
            thumbnail_source_path,
            out_path
        )

        return out_path
```

Or, as a diff:

```diff,python
     def get_thumbnail(self):
         """
         The thumbnail of a PDF is just a 500px wide image of the first page.
         """
 
         out_path = os.path.join(self.tempdir, "convert.png")
         gs_out_path = os.path.join(self.tempdir, "gs_out.png")
 
-        # Run convert to get a decent thumbnail
-
+        # Extract the first PDF page as a PNG using Ghostscript, if available, to use as a thumbnail.
         # https://github.com/danielquinn/paperless/issues/447
-        # call gs first
-        cmd = ["gs", "-q", "-sDEVICE=pngalpha",
-               "-o", gs_out_path, self.document_path]
-        if not subprocess.Popen(cmd).wait() == 0:
-            raise ParseError("Thumbnail (gs) failed at {}".format(cmd))
-        # then run convert on the output from gs
+        try:
+            cmd = [
+                self.GHOSTSCRIPT,
+                "-q",
+                "-sDEVICE=pngalpha",
+                "-o", gs_out_path,
+                self.document_path
+            ]
+            if not subprocess.Popen(cmd).wait() == 0:
+                raise ParseError("Thumbnail (gs) failed at {}".format(cmd))
+            thumbnail_source_path = gs_out_path
+        except FileNotFoundError:
+            thumbnail_source_path = "{}[0]".format(self.document_path)
+
+        # Run convert to get a decent thumbnail
         run_convert(
             self.CONVERT,
             "-scale", "500x5000",
             "-alpha", "remove",
-            "gs_out_path",
+            thumbnail_source_path,
             out_path
         )
 
         return out_path
```

src/paperless_tesseract/parsers.py
```diff
@@ -45,13 +45,22 @@ def get_thumbnail(self):
...
+               "-o", gs_out_path, self.document_path]
+        if not subprocess.Popen(cmd).wait() == 0:
+            raise ParseError("Thumbnail (gs) failed at {}".format(cmd))
+        # then run convert on the output from gs
         run_convert(
             self.CONVERT,
             "-scale", "500x5000",
             "-alpha", "remove",
-            "{}[0]".format(self.document_path),
+            "gs_out_path",
```
pitkley: I think quoting this is wrong and will not work, i.e. it should be:

```suggestion
            gs_out_path,
```

src/paperless_tesseract/parsers.py
JensPfeifle: I like the idea of a fallback. But in this case if ghostscript is not available we depend on imagemagick, which may or may not fail on parsing.

What if we make gs a requirement and make sure that it is available by adding it to the binaries check in [checks.py](https://github.com/the-paperless-project/paperless/blob/6d8e4ad6bf92094149e2335a8f30cb976475fe70/src/paperless/checks.py#L70)?

src/paperless_tesseract/parsers.py
JensPfeifle: Dumb mistake.. thanks.
Interestingly enough, it still worked when I tested it..

JensPfeifle:
There is a similar call of convert in [a Django migrations file](https://github.com/the-paperless-project/paperless/blob/6d8e4ad6bf92094149e2335a8f30cb976475fe70/src/documents/migrations/0012_auto_20160305_0040.py#L86). Does it make sense to apply a fix here as well?

src/paperless_tesseract/parsers.py
erikarvstedt: Because ImageMagick is using `Ghostscript` internally for its PDF operations, Ghostscript is already a strict dependency of Paperless (although not explicitly checked in `checky.py`).
So that `try: ... except:` is unneeded.
Actually, we should try the fast single pass `"{}[0]"` approach first, and then fall back to the slower `Ghostscript + ImageMagick` workaround.

erikarvstedt:
This fix is specific to PDFs, but the document can also be an (multi-page) image.
So please keep the old approach that works universally for all image formats and, when that fails, fall back to the Ghostscript workaround for PDFs.

JensPfeifle:
Thanks for the input so far. I have added the appropriate settings and set it up so that the workaround applies only if `consume` fails. In that case, a warning is written to the log. PNG and JPG consume is working as well.

**Consuming the failing example document posted in [#447](https://github.com/the-paperless-project/paperless/issues/447#issuecomment-449980883)**:
image

Is `warning` appropriate? Or rather `info`?

src/paperless_tesseract/parsers.py
JensPfeifle: Resolved by https://github.com/the-paperless-project/paperless/pull/494/commits/dd0bbc48286addb0e68012a0e3ada8d3c576e24c

JensPfeifle: Resolved by https://github.com/the-paperless-project/paperless/pull/494/commits/448e5c201bbf4d195ea657d519d71fca34222e08

pitkley:
I am going to close and reopen this PR, since that [seems to fix Travis](https://github.com/the-paperless-project/paperless/issues/512#issuecomment-469050805) (thanks for discovering this easy fix, @stgarf 👍).

JensPfeifle:
Thanks Pit  :ok_hand:

src/paperless_tesseract/parsers.py
```diff
@@ -47,13 +48,38 @@ def get_thumbnail(self):
...
+                   "-q",
+                   "-sDEVICE=pngalpha",
+                   "-o", gs_out_path,
+                   self.document_path]
+            if not subprocess.Popen(cmd).wait() == 0:
+                raise ParseError("Thumbnail (gs) failed at {}".format(cmd))
+            # then run convert on the output from gs
+            run_convert(
+                self.CONVERT,
+                "-scale", "500x5000",
```
MasterofJOKers: I'm not sure whether we should move this into a variable, as this is used in two places and one of them might be forgotten if it's ever changed. Could be done like this:
```python
from functools import partial

convert = partial(run_convert,
    self.CONVERT,
    "-scale", "500x5000",
    "-alpha", "remove")
```
and then use it like so
```python
convert("{}[0]".format(self.document_path), out_path)
```
and
```python
convert(gs_out_path, out_path)
```

@pitkley Any comments?

JensPfeifle: My thoughts: 
I generally agree, but the two calls to `run_convert` are very close together - in one try-except block. And I would expect that anyone making (or reviewing) changes will at least read the whole function.
I have no problem doing it with `partial`, but to me the additional layer of abstraction makes it harder to understand at first glance what the code is doing.

MasterofJOKers: Fair enough.

ddddavidmartin:
I'm happy to see this merged in! Is there any chance someone could create a release that includes these changes? Then it would update the docker image on dockerhub as well. Thanks!

stgarf:
@pitkley Do you have those permissions possibly?

pitkley:
@stgarf I don't. :slightly_frowning_face:. The way I see it the issue with Docker right now [is it not picking up the moved repository](https://github.com/the-paperless-project/paperless/issues/470#issuecomment-466815872). Otherwise it would build from `master` automatically. And the "old" repository [`danielquinn/paperless`](https://hub.docker.com/r/danielquinn/paperless) doesn't receive updates either...

I guess this is only something @danielquinn can fix, I'll respond to his comment in issue #470.

Issue #495, opened 2019-01-29
Logging improvments (emails & sentry)


ovv:
I plan to work on improving the current logging setup when I have time. First [sentry](https://sentry.io/welcome) and I'll probably try to look into adding a configuration option for sending stack-traces by emails.

syntonym:
I would personally prefer to have the logging in the same file as other configuration.

Does sentry send diagnostic information on crashes to the developers? As paperless can possible deal with very sensitive data (bills, taxes, healthcare) I think we should be careful about user acceptance.

PR #496, opened 2019-01-30
WIP: Logging configuration

ovv:
Putting it out here to get feedback on the approach.

I figured adding a file to configure the whole of logging would be better than individual option (See #495 ).
I'm not sure how to best name the file, creating a `paperless` directory and simply naming it `logging.yml` seems fine but it changes where the configuration should be located.

I still need to do some more tests and write some documentation

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,7 +15,7 @@ django-filter = "*"
 djangorestframework = "*"
 factory-boy = "*"
 filemagic = "*"
-fuzzywuzzy = {extras = ["speedup"], version = "==0.15.0"}
+fuzzywuzzy = {extras = ["speedup"],version = "==0.15.0"}
 gunicorn = "*"
 inotify-simple = "*"
 langdetect = "*"
@@ -36,6 +36,7 @@ pytest-env = "*"
 pytest-xdist = "*"
 psycopg2 = "*"
 djangoql = "*"
+pyyaml = "*"
 
 [dev-packages]
 ipython = "*"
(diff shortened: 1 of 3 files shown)


danielquinn:
This is all good stuff, and I just have a few notes:

1. I like the move to use YAML for configuration.  Especially in the case of logging, this is a natural choice.  My only concern though is that we're introducing another config file in *another* format, which isn't very user-friendly.  Paperless has always followed the [12-factor](https://12factor.net/) methodology (configuration only through environment variables) to keep things portable, and adding this new file (and format) introduces a requirement for the existence of a readable file.  If we're going to introduce this requirement, we should be conscious of the limitations it'll introduce in terms of portability between setups (docker vs bare metal vs some other implementation).
2. The use of Sentry is a great idea, but it's probably best to make use of their `raven` library rather than rolling our own integration.  Sentry has some [pretty good docs](https://docs.sentry.io/clients/python/integrations/django/) on this subject, but the TL;DR is that you need to add their library to `INSTALLED_APPS` which could be done in `settings.py` dynamically after detecting the preferred configuration.

danielquinn:
Further to :one: above, would it be altogether crazy to set an environment variable to a JSON value that would represent the logging config?  It'd be ugly, but at least more portable.  Something like:

```
PAPERLESS_LOGGING={"version": 1, "disable_existing_loggers": false, "handlers": {"consumer": {"class": "documents.loggers.PaperlessLogger"}}, "loggers": {"documents": {"handlers": ["consumer"], "level": "INFO"}}}
```

PR #498, opened 2019-02-03
WIP: Improved tests for consumption process

syntonym:
The consumption process is a critical path for paperless, as crashes are hard to detect for the user, but render the paperless instance useless until the user detects the crash. Thus it is critical to have a good test suite for the consumption process.

The tests so for focus on the "happy path", i.e. paperless works correctly if everything works as expected, but there are currently no tests for the case that external sources introduce problems. For example the internet connection could go down when paperless tries to pull for mails or some permissions change after reading the file. For this pull request I focused on paperless-external sources of errors.

To make this more systematic I reviewed the file `documents/consumer.py` for function calls that communicate with the external world i.e. internet requests and file operations. I assume that all these functions live in a module imported at the top of the file or are a builtin. Of all modules imported only `os` seems to be able to fail due to external sources. Of the builtins used only `open` seems to be of interest. (These assumptions are not correct, as at least parsers returned from `document_consumer_declaration` can surely fail due to external sources. But lets first start somewhere and then work from that.)

Just looking at `open` there are four cases which can fail and where exceptions are not handled. Exception handling is also not done in any calling code, so if any of these four calls fail, the consumer process will crash. To use a TDD approach (and check if my understanding that an exception will actually produce a crash) I added some tests that make open fail. If open fails, it doesn't have to be the first call to open that fails. It could happen that paperless already started the consumption process and then a different process e.g. deletes the file paperless is currently consuming. Some errors are surely less probable than others (e.g. permission changed at the end of consumption seems pretty unlikely), but at least in the consumption process I think better safe than sorry is the better attitude.

This is very much work in progress, e.g. currently the parser tries to open the file which is not there (and then fails but does not crash as expected), which needs to be mocked out. I wanted to get some feedback if a) this approach seems sensible and b) if the way I intend to test this makes sense. After I'm done it sure makes sense for others to look over if I missed some modules that could introduce external errors.

### DIFF
--- a/src/documents/tests/test_consumer.py
+++ b/src/documents/tests/test_consumer.py
@@ -1,10 +1,23 @@
 from django.test import TestCase
 from unittest import mock
 from tempfile import TemporaryDirectory
+from io import StringIO
+import pytest
 
 from ..consumer import Consumer
 from ..models import FileInfo, Tag
 
+
+def error_on_call_n(error, n):
+    """Returns an object which raises an error on the n-th call"""
+    count = [0]
+    def error_object(*args, **kwargs):
+        if count[0] == n:
+            raise error
+        else:
+            count[0] += 1
+        return mock.DEFAULT
+    return error_object
 
 class TestConsumer(TestCase):
 
@@ -16,6 +29,33 @@ class TestConsumer(TestCase):
             self._get_consumer()._get_parser_class("doc.pdf"),
             self.DummyParser
         )
+
+    def test_open_errors(self):
+        """
+        Test that exceptions on open do not crash the consumer
+
+        It is not enough to test the consumption process with `open` throwing
+        an error everytime.  The first open call could succeed, but then the
+        file gets moved or permission changed by another process (e.g. user
+        manually intervening or a second paperless process).  Thus we need to
+        check what happens when only the n-th call fails.
+        """
+        
+        # open can throw PermissionError or FileNotFoundError
+        for error in [PermissionError(), FileNotFoundError()]:
+            # Check exception on n-th open call
+            for n in range(0, 4):
+                # Properly mock open
+                mockfile = mock.Mock()
+                mockfile_context_manager = mock.MagicMock()
+                # Return "I'm a pdf" on read()
+                mockfile_context_manager.read.return_value = StringIO("I'm a pdf")
+                mockfile.__enter__ = mockfile_context_manager
+                mockfile.__exit__ = mock.MagicMock()
+                with mock.patch("builtins.open", side_effect=error_on_call_n(error, n)) as mockopen:
+                    mockopen.return_value = mockfile
+                    with TemporaryDirectory() as tmpdir:
+                        consumer = Consumer(tmpdir).try_consume_file("testfile.pdf")
 
     @mock.patch("documents.consumer.os.makedirs")
     @mock.patch("documents.consumer.os.path.exists", return_value=True)


src/documents/tests/test_consumer.py
```diff
@@ -17,6 +30,33 @@ def test__get_parser_class_1_parser(self):
...
+        an error everytime.  The first open call could succeed, but then the
+        file gets moved or permission changed by another process (e.g. user
+        manually intervening or a second paperless process).  Thus we need to
+        check what happens when only the n-th call fails.
+        """
+        
+        # open can throw PermissionError or FileNotFoundError
+        for error in [PermissionError(), FileNotFoundError()]:
+            # Check exception on n-th open call
+            for n in range(0, 4):
```
ovv: I'm not sure what test runner paperless uses but there should be a way to have those two `for` loops be parameters.
See [pytest documentation](https://docs.pytest.org/en/latest/parametrize.html) for what I mean.

ovv (changes requested):
I only gave a quick look through the PR but here is some feedback. In any cases more tests are always useful :+1:

src/documents/tests/test_consumer.py
```diff
@@ -1,11 +1,24 @@
...
+from io import StringIO
+import pytest
 
 from ..consumer import Consumer
 from ..models import FileInfo, Tag
 
 
+def error_on_call_n(error, n):
+    """Returns an object which raises an error on the n-th call"""
+    count = [0]
```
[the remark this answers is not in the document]
syntonym: I should have really put a comment there. Python scoping rules are really weird. The code without the list fails:

def error_on_call_n(error, n):
    count = 0
    def error_object(*args, **kwargs):
        if count == n:
            raise error
        else:
            count += 1
    return error_object

with an UnboundLocalError: local variable 'count' referenced before assignment. Because of the assignment `count += 1` python understands `count` as a local variable (with the scope of the function `error_object`), but when checking the if condition `count == n` the local variable `count` is not existent. Changing count to a list changes count[0] += 1 from an "variable assignment" to an "operation on an object" so python does not understand count as a local variable but ("correctly") looks it up (so that it finds it in the scope of `error_on_call_n`).

The "correct" solution is apparently to tell python that count is nonlocal:

def error_on_call_n(error, n):
    count = 0
    def error_object(*args, **kwargs):
        nonlocal count
        if count == n:
            raise error
        else:
            count += 1
    return error_object

but to be honest I have never seen a `nonlocal` in the wild. While the list solution is confusing why exactly it was chosen, I think every python developer understands what is happening. With `nonlocal` I would assume that more people would have to lookup what that even is.

Should I change that to the `nonlocal` solution?

src/documents/tests/test_consumer.py
syntonym: As far as I understand currently the tests are written in the `unittest.TestCase` style and pytest [does not support parametrization of tests written in that style](https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses). Is pytest "given" and can I write tests (with parameterization) that will only work with pytest but e.g. not the "standard library unittest" module?

src/documents/tests/test_consumer.py
```diff
@@ -17,6 +30,33 @@ def test__get_parser_class_1_parser(self):
...
+            # Check exception on n-th open call
+            for n in range(0, 4):
+                # Properly mock open
+                mockfile = mock.Mock()
+                mockfile_context_manager = mock.MagicMock()
+                # Return "I'm a pdf" on read()
+                mockfile_context_manager.read.return_value = StringIO("I'm a pdf")
+                mockfile.__enter__ = mockfile_context_manager
+                mockfile.__exit__ = mock.MagicMock()
+                with mock.patch("builtins.open", side_effect=error_on_call_n(error, n)) as mockopen:
```
danielquinn: If I understand this correctly, you've counted the number of times we use `open()` in the consumer path and are looping over each making sure we handle `PermissionError` and `FileNotFoundError`.  Given that it's entirely likely that future iterations of the code may move these uses of `open()` around, and even add/remove some, I would think that this makes the test unnecessarily brittle.  Wouldn't it be better to write separate unit tests for each method with failing uses of `open()` expressly defined?  At least then if we change things, it's easy to see why a test failed and fix it.

danielquinn (commented):
I like the approach here: more tests along the not-so-happy-path are always welcome.  I think it's a good start.

src/documents/tests/test_consumer.py
syntonym: Do you mean "We don't want to use PyTest style"? Because pytest is currently a [dependency](https://github.com/the-paperless-project/paperless/blob/master/requirements.txt#L54) of paperless. I'm not sure where/how tests are actually run, but I think it's also done [by pytest (by travis)](https://github.com/the-paperless-project/paperless/blob/master/.travis.yml#L37).

src/documents/tests/test_consumer.py
syntonym: I think more of this testcase as an integration test, so that `consume()` is tested as close as possible to production as possible. Optimally it should never silently fail. When unittesting/testing subfunctions separately it is a) similar easy to forget to test a subfunction b) disregarding interactions between subfunctions. But I agree that this current design is brittle.

I see four ways forward:

1) Go full integration testing, run paperless in some kind of VM and let it actually consume stuff. Pro: Closest to production deployment Cons: Might be generally hard (syncronizing failing conditions, running in vm, proper sampling of input and conditions)
2) Keep the current approach but try to make it less brittle, increase the number of opens tested, or maybe automatically count how often open is called. Add similar functions that might fail. Pros: If done correctly should be faster and easier than 1, but have similar rates of catching bugs. Cons: Brittle, limited sampling of failing conditions, might produce weird failing messages.
3) Go more direction unittesting, test seperate functions. Pros: Should be less brittle (regarding refactoring) Cons: Does not capture interactions between units, is "far away from production conditions".
4) Cover consume() in a `try... catch` and display any such error messages prominently to the user. Make it easy for the user to a) see that one document failed consumption (no silent dropping of documents) b) post error messages to the issue tracker. Wasn't there a discussion on automatically sending logs somewhere? Pros: Captures all bugs in production. User gets feedback when something goes wrong, in the other cases not identified bugs could still lead to silently dropping documents. Cons: Captures bugs only in production, does not actually identify bugs when developing.

When I wrote the PR I was favoring 2. I've sitting a different version of the PR in a local repository, where I rewrote the tests to use pytests advanced features. But I actually couldn't reproduce some crashes which I thought should be there: `open` throwing errors did not crash consume(), but I couldn't identify why.

Currently I'd probably go for 4.

PR #501, opened 2019-02-06
* Docs: Explain how to view consumer logs

bmsleight:


docs/troubleshooting.rst
```diff
@@ -73,3 +73,14 @@ For more information on this and situations like it, you should take a look
...
+
+.. _troubleshooting-consumerlogs
+
+
+Viewing Consumer logs
+---------------------
+
+.. code:: bash
+
+    $ sudo journalctl -u paperless-consumer
```
pitkley: I'm not quite sure to what installation method this is applicable. At least for my setup, I have to check the logs through Docker.

Can you:

* Describe what command to use, depending on the type of installation. (You don't necessarily have to cover all of them, but at least make clear to which method which command applies.)
* Add a sentence on what the consumer logs show, i.e. why a troubleshooter would want to look into them.
* Maybe add a bit of example output if you think it makes sense -- I'm not quite sure myself.

PR #515, opened 2019-02-27
define export in both services

LunkSnee:
If export isn't defined in both services, docker creates a volume for it under webserver.
Adding export under both consolidates the volumes and stops docker from creating a second export volume.


docker-compose.yml.example
```diff
@@ -20,6 +20,12 @@ services:
             # You have to adapt the local path you want the consumption
             # directory to mount to by modifying the part before the ':'.
             - ./consume:/consume
+						# Likewise, you can add a local path to mount a directory for
```
danielquinn: This line is lacking 1 space of indentation.  Will that work in YAML?

PR #527, opened 2019-03-17
Strip the thumbnails

sbrunner:
I regularly get thumbnails like this

then I purpose to strip on the content.


danielquinn:
I didn't even know this was a thing you could do!

sbrunner:
Thanks :-)

PR #530, opened 2019-03-29
Fix issue where additional tesseract langages weren't installed by docker-entrypoint.sh

### LINKED ISSUE #528: Consumer can't OCR Swedish document
Yes, Liek that title says, I drop a pdf in Swedish, and it can't be added.

docker-compose.env has the line
"PAPERLESS_OCR_LANGUAGES=swe eng"

and so does the web server via the file docker-compose.yml

The log says;

"OCRing the document
Parsing for eng
Parsing for swe
PARSE FAILURE for /consume/filename.pdf: The guessed language (sv) is not available in this instance of Tesseract."

"sv" = "Swe" = Swedish, in case you were wondering...

schwabelbauch:
fixes #528

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -82,7 +82,7 @@ install_languages() {
         if apk info -e "$pkg" > /dev/null 2>&1; then
             continue
         fi
-        if ! apk info "$pkg" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


ahyear:
it seams that it's not ok after all :
in /usr/share/tessdata/ there is missing the eng files... i'm not sure why...

PR #532, opened 2019-04-01
Proposal to deploy Docker image on each commit on master

sbrunner:


stgarf:
@sbrunner It might make sense to push a docker image to DockerHub only on tagged releases to ensure we're only pushing quality containers based on tagged releases... (instead of each commit to master which may or may not contain a variety of in-progress work and bugs).

.travis.yml
```diff
@@ -12,15 +12,18 @@ matrix:
         - python: "3.5"
         - python: "3.6"
         - python: "3.7-dev"
-        - env: BUILD_DOCKER=1
+        - env:
+            - BUILD_DOCKER=1
+            - DOCKER_USERNAME=sbrunner
+            - DOCKER_PASSWORD=<replace by `travis encrypt DOCKER_PASSWORD=real-password`>
```
ahyear: I am not sure it's such a good idea to have a docker_password (even encrypted) in a visible repository visible file...

stgarf: https://docs.travis-ci.com/user/encryption-keys/

```$ travis encrypt --com DOCKER_PASSWORD="secretvalue"```

Comment (ahyear) [reply .travis.yml]: ok I guess it's a non issue then...

**Comment by sbrunner** [comment]
@stgarf Yes, I will work on that :-)

**Comment by ahyear** [comment]
is it possible, with a parametre in docker-compose.env (for exemple) to let the user choose between last tag and master?

**Comment by sbrunner** [comment]
@stgarf I change it and also test it as most as possible without a real password ...

**Comment by sbrunner** [comment]
@ahyear
It's not possible to do that in the `docker-compose.env` because he contains only environment variable for the container, to do that we should create a `.env` file.
And finally for me it should be done in another pull request...

**Comment by stgarf** [comment]
@ahyear 
> is it possible, with a parametre in docker-compose.env (for exemple) to let the user choose between last tag and master?

To do what you're asking you'd instead replace the `build: ./` [here](https://github.com/the-paperless-project/paperless/blob/1245297939fb90848b121c3a41f1f749513cdaee/docker-compose.yml.example#L5) and [here](https://github.com/the-paperless-project/paperless/blob/1245297939fb90848b121c3a41f1f749513cdaee/docker-compose.yml.example#L33) with something à la `image: <docker-image-you/want-to-run>:<tag-you-want>` in your `docker-compose.yaml`.

**Comment by sbrunner** [comment]
What can be the next for this pull request?

**Comment by MasterofJOKers** [comment]
If we merge this, the deploy step will not be active, right? I assume it's not, because it should fail without a password and travis shows no failure for this PR.

So after merging this, someone™ would have to change the username, encrypt and set the password and activate the deploy step. Who would do that and how do you activate that step?

**Comment by sbrunner** [comment]
@MasterofJOKers You just needs to fill the DOCKER_USERNAME and DOCKER_PASSWORD, it don't failed because only tags and the master branch is published...

File: .travis.yml [hunk 2]
```diff
@@ -12,15 +12,18 @@ matrix:
...
           services:
             - docker
           before_install:
             - true
           install:
             - true
           script:
-            - docker build -t paperless .
+            - docker build --tag=the-papaerless-project/paperless .
```
MasterofJOKers: There's a typo here. "papaer"

.travis.yml
MasterofJOKers: To be honest, I'd prefer this to be empty/commented and the script to check if the variables are set instead of it blindly using them. Especially, since we cannot set them in this PR.

MasterofJOKers:
> @MasterofJOKers You just needs to fill the DOCKER_USERNAME and DOCKER_PASSWORD, it don't failed because only tags and the master branch is published...

Ah. So it will only fail for master, once we merge it, but since PRs aren't included, travis would still be green for them. I feared, we would end up not being able to merge anything afterwards.

Still, whoever has the appropriate rights on their account to allow pushing into docker should merge this PR, because she/he would have to create a commit right afterwards.

.travis.yml
sbrunner: Done :-)

.travis.yml
sbrunner: OK, done, then if there are missing I exit without error :-)

sbrunner:
Thanks @MasterofJOKers for your review, I just the required changes :-), Your right it's better ton don't break the master :-)

sbrunner:
UP ...

sbrunner:
Thanks for the approved But how can we merge in configure the Docker deployment?

sbrunner:
Thanks @ahyear  :-)

danielquinn:
There's a discussion about this (and the future of how we handle Docker integration) going on [here](https://github.com/the-paperless-project/paperless/issues/470#issuecomment-492265549).

PR #536, opened 2019-04-10
install default tesseract eng configuration in docker

ahyear:


MasterofJOKers:
There doesn't seem to be a package like that [in the repos](https://pkgs.alpinelinux.org/packages?name=tesseract-ocr-data-eng&branch=v3.8&arch=x86_64). Here are the [available packages](https://pkgs.alpinelinux.org/packages?name=tesseract-ocr-data-*&branch=v3.8&arch=x86_64).

Also the `tesseract-ocr` package contains data for the english language, as it seems: [link to package content](https://pkgs.alpinelinux.org/contents?branch=v3.8&name=tesseract-ocr&arch=x86_64&repo=community).

So I assume the problem comes from somewhere else and would prefer closing this PR.

ahyear:
To be sure I tried with the previous commit and got only English and after the commit all but eng... Not sure what to think ...

PR #539, opened 2019-05-05
Warn if consume directory contains subdirectories

bauerj:
Currently, the consumer ignores files in subdirectories of the consume directory.

This changes the behaviour to find files recursively instead.

While switching to Paperless, I copied my old directory structure with subdirectories for each year into the consume directory. The consumer didn't pick up anything which got me a bit confused.

### DIFF
--- a/src/documents/consumer.py
+++ b/src/documents/consumer.py
@@ -94,6 +94,11 @@ class Consumer:
                     ignored_files.append(file)
                 else:
                     files.append(file)
+            else:
+                self.logger.warning(
+                    "Skipping %s as it is not a file",
+                    entry.path
+                )
 
         if not files:
             return
--- a/src/documents/management/commands/document_consumer.py
+++ b/src/documents/management/commands/document_consumer.py
@@ -26,6 +26,7 @@ class Command(BaseCommand):
     def __init__(self, *args, **kwargs):
 
         self.verbosity = 0
+        self.logger = logging.getLogger(__name__)
 
         self.file_consumer = None
         self.mail_fetcher = None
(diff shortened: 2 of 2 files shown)


erikarvstedt:
This won't work when `inotify` is enabled, which is the default on Linux.
An equivalent inotify implementation is non-trivial, or even impossible for many sub-directories due to `fs.inotify.max_user_watches` limits. But it may be eased with the upcoming [`fanotify` improvements](https://github.com/amir73il/fsnotify-utils/wiki/Super-block-root-watch) on Linux.

I don't think this feature is worth the extra implementation complexity. It also makes it harder to port inotify-like file system watching to other operating systems.

bauerj:
That makes sense, thank you! 

In that case, do you think it should show a warning if it encounters a directory there?

erikarvstedt:
Yeah, that would be fine, like `Warning: Skipping directory 'mydir'`.

erikarvstedt:
Could you add the same warning to the [inotify part](https://github.com/the-paperless-project/paperless/blob/351809b8a93afc44167e16303ee8d74579aa3090/src/documents/management/commands/document_consumer.py#L139)?

bauerj:
Yes, thanks for pointing that out!

erikarvstedt:
Note that the CI is still failing:
```
./documents/management/commands/document_consumer.py:142:80: E501 line too long (93 > 79 characters)
The command "pycodestyle" exited with 1.
```

bauerj:
Sorry about that.

MasterofJOKers:
I agree with @danielquinn: we should use a generic `else` instead of `elif` with `is_dir()`. I can think of a case, where a user would add a named-pipe into the directory, expecting to be able to stream output from another program into the consumer. With the current change, that user would run into the same debug-session you did originally.

According to the [the docs](https://docs.python.org/3/library/os.html?highlight=os#os.DirEntry.is_file) and [the docs](https://docs.python.org/3/library/os.path.html#os.path.isfile), symlinks to files are actually covered.

bauerj:
@MasterofJOKers I agree with your reasoning. I have changed the PR now to the more generic case.

PR #542, opened 2019-05-18
Allow configuring transformations to be applied to the filename before

grembo:
parsing. The motivation was that files produced by a Brother scanner
wouldn't match paperless' expectations. At most one transformation
is applied (first matching). It won't affect the filename on disk.

This is generic enough so that it is useful for various purposes.
In my case it allows me to use the different hardware buttons on
the scanner to use different profiles, feeding one instance of
paperless with documents of multiple entities and tagging them
accordingly.

Example:

`PAPERLESS_FILENAME_PARSE_TRANSFORMS=[{"pattern":"^([a-z]+)_(\\d{8})_(\\d{6})_([0-9]+)\\.", "repl":"\\2\\3Z - \\4 - \\1."},{"pattern":"^([a-z]+)_([0-9]+)\\.", "repl":" - \\2 - \\1."}]`

### DIFF
--- a/docs/guesswork.rst
+++ b/docs/guesswork.rst
@@ -54,6 +54,34 @@ filename as described above.
 
 .. _dateparser: https://github.com/scrapinghub/dateparser/blob/v0.7.0/docs/usage.rst#settings
 
+Transforming filenames for parsing
+----------------------------------
+Some devices can't produce filenames that can be parsed by the default
+parser. By configuring the option ``PAPERLESS_FILENAME_PARSE_TRANSFORMS`` in
+``paperless.conf`` one can add transformations that are applied to the filename
+before it's parsed.
+
+The option contains a list of dictionaries of regular expressions (key:
+``pattern``) and replacements (key: ``repl``) in JSON format, which are
+applied in order by passing them to ``re.subn``. Transformation stops
+after the first match, so at most one transformation is applied. The general
+syntax is
+
+.. code:: python
+
+   [{"pattern":"pattern1", "repl":"repl1"}, {"pattern":"pattern2", "repl":"repl2"}, ..., {"pattern":"patternN", "repl":"replN"}]
+
+The example below is for a Brother ADS-2400N, a scanner that allows
+different names to different hardware buttons (useful for handling
+multiple entities in one instance), but insists on adding ``_<count>``
+to the filename.
+
+.. code:: python
+
+   # Brother profile configuration, support "Name_Date_Count" (the default
+   # setting) and "Name_Count" (use "Name" as tag and "Count" as title).
+   PAPERLESS_FILENAME_PARSE_TRANSFORMS=[{"pattern":"^([a-z]+)_(\\d{8})_(\\d{6})_([0-9]+)\\.", "repl":"\\2\\3Z - \\4 - \\1."}, {"pattern":"^([a-z]+)_([0-9]+)\\.", "repl":" - \\2 - \\1."}]
+
 .. _guesswork-content:
 
 Reading the Document Contents
--- a/src/documents/models.py
+++ b/src/documents/models.py
@@ -484,11 +484,15 @@ class FileInfo:
         """
 
         filename = os.path.basename(path)
+
+        # Mutate filename in-place before parsing its components
+        # by applying at most one of the configured transformations.
         for (pattern, repl) in settings.FILENAME_PARSE_TRANSFORMS:
             (filename, count) = pattern.subn(repl, filename)
             if count:
                 break
 
+        # Parse filename components.
         for regex in cls.REGEXES.values():
             m = regex.match(filename)
             if m:
--- a/src/documents/models.py
+++ b/src/documents/models.py
@@ -483,8 +483,14 @@ class FileInfo:
           "<title>.<suffix>"
         """
 
+        filename = os.path.basename(path)
+        for (pattern, repl) in settings.FILENAME_PARSE_TRANSFORMS:
+            (filename, count) = pattern.subn(repl, filename)
+            if count:
+                break
+
         for regex in cls.REGEXES.values():
-            m = regex.match(os.path.basename(path))
+            m = regex.match(filename)
             if m:
                 properties = m.groupdict()
                 cls._mangle_property(properties, "created")
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -10,7 +10,9 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/1.10/ref/settings/
 """
 
+import json
 import os
+import re
 
 from dotenv import load_dotenv
 
@@ -316,6 +318,15 @@ FY_END = os.getenv("PAPERLESS_FINANCIAL_YEAR_END")
 # Specify the default date order (for autodetected dates)
 DATE_ORDER = os.getenv("PAPERLESS_DATE_ORDER", "DMY")
 FILENAME_DATE_ORDER = os.getenv("PAPERLESS_FILENAME_DATE_ORDER")
+
+# Transformations applied before filename parsing
+FILENAME_PARSE_TRANSFORMS = []
+_filename_parse_transforms = os.getenv("PAPERLESS_FILENAME_PARSE_TRANSFORMS")
+if _filename_parse_transforms:
+    FILENAME_PARSE_TRANSFORMS = [(
+            re.compile(t["pattern"]), t["repl"])
+        for t in json.loads(_filename_parse_transforms)
+    ]
 
 # Specify for how many years a correspondent is considered recent. Recent
 # correspondents will be shown in a separate "Recent correspondents" filter as
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -321,12 +321,8 @@ FILENAME_DATE_ORDER = os.getenv("PAPERLESS_FILENAME_DATE_ORDER")
 
 # Transformations applied before filename parsing
 FILENAME_PARSE_TRANSFORMS = []
-_filename_parse_transforms = os.getenv("PAPERLESS_FILENAME_PARSE_TRANSFORMS")
-if _filename_parse_transforms:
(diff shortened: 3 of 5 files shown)


grembo:
Note that I extended the pull request to add a protocol to CORS headers, otherwise the following error is generated with recent versions:

```
SystemCheckError: System check identified some issues:

ERRORS:
?: (corsheaders.E013) Origin 'localhost:8080' in CORS_ORIGIN_WHITELIST is missing scheme or netloc
	HINT: Add a scheme (e.g. https://) or netloc (e.g. example.com).

```

docs/guesswork.rst
```diff
@@ -54,6 +54,34 @@ filename as described above.
...
+
+The option contains a list of dictionaries of regular expressions (key:
+``pattern``) and replacements (key: ``repl``) in JSON format, which are
+applied in order by passing them to ``re.sub``. Transformation stops
+after the first match, so at most one transformation is applied. The general
+syntax is
+
+.. code:: python
+
+   [{"pattern":"pattern1", "reply":"repl1"}, {"pattern":"pattern2", "reply":"repl2"}, ..., {"pattern":"patternN", "reply":"replN"}]
```
pitkley:
```suggestion
   [{"pattern":"pattern1", "repl":"repl1"}, {"pattern":"pattern2", "repl":"repl2"}, ..., {"pattern":"patternN", "repl":"replN"}]
```

src/paperless/settings.py
```diff
@@ -102,7 +104,7 @@ def __get_boolean(key, default="NO"):
 ]
 
 # We allow CORS from localhost:8080
-CORS_ORIGIN_WHITELIST = tuple(os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "localhost:8080").split(","))
+CORS_ORIGIN_WHITELIST = tuple(os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "http://localhost:8080").split(","))
```
pitkley: This [only applies for django-cors-headers version 3](https://github.com/adamchainz/django-cors-headers/blob/master/HISTORY.rst#300-2019-05-10), and the Paperless [lockfile still references 2.4.0](https://github.com/the-paperless-project/paperless/blob/master/Pipfile.lock#L126-L133). So unless you have upgraded the package on your installation, this should not fail.

What I'm getting at: we should either update both the plugin and the whitelist, or none. Also: if you decide to update the plugin to v3, can you do that in a separate PR?

src/documents/models.py
```diff
@@ -483,8 +483,14 @@ def from_path(cls, path):
           "<title>.<suffix>"
         """
 
+        filename = os.path.basename(path)
+        for transform in settings.FILENAME_PARSE_TRANSFORMS:
+            (filename, count) = transform[0].subn(transform[1], filename)
```
pitkley: For clarity I'd suggest to destructure the `transform` tuple:

```python
for (pattern, repl) in settings.FILENAME_PARSE_TRANSFORMS:
    (filename, count) = pattern.subn(repl, filename)
```

src/paperless/settings.py
```diff
@@ -317,6 +319,15 @@ def __get_boolean(key, default="NO"):
 DATE_ORDER = os.getenv("PAPERLESS_DATE_ORDER", "DMY")
 FILENAME_DATE_ORDER = os.getenv("PAPERLESS_FILENAME_DATE_ORDER")
 
+# Transformations applied before filename parsing
+FILENAME_PARSE_TRANSFORMS = []
+_filename_parse_transforms = os.getenv("PAPERLESS_FILENAME_PARSE_TRANSFORMS")
+if _filename_parse_transforms:
+    FILENAME_PARSE_TRANSFORMS = [(
+            re.compile(t["pattern"], flags = re.IGNORECASE), t["repl"])
```
pitkley: I'm not sure if ignoring case should be the default given it can be activated through the in-line flag `(?i)` but not de-activated through such a flag. I guess people using regular expressions probably know just enough to understand they are case-sensitive by default, although I understand where this is coming from and I'm not strictly opposed.

Either way, it should definitely be documented! Can you add a note about this to the guesswork page?

src/paperless/settings.py
grembo: I stumbled over this as I maintain paperless' FreeBSD port which is rolling release - so the package updated to a newer version there. Will adding http:// here break anything on django-cors-headers 2.4.0? (I can of course also update the plugin to three).

docs/guesswork.rst
grembo: Correct

src/paperless/settings.py
grembo: I'll back out that change for now and open a new pull request later, as suggested

src/documents/models.py
grembo: I like that better as well, thanks for the suggestion

src/paperless/settings.py
pitkley: Ah, I didn't realize that the scheme was supported, but simply optional before. That definitely makes changing this a no-brainer, even if you don't update the plugin. A separate PR for this sounds great 👍

src/paperless/settings.py
grembo: I'll remove ignorecase here, usually devices only emit one format so one hardly would need case insensitive matches anyway.

src/paperless/settings.py
grembo: Actually, I'm not entirely sure if 2.4 can handle schemes, so I'll go with what you suggested once we're done here (separate PR and update plugin to 3)

src/documents/models.py
```diff
@@ -483,8 +483,14 @@ def from_path(cls, path):
           "<title>.<suffix>"
         """
 
+        filename = os.path.basename(path)
```
danielquinn: This is crafty, I like it.  Can you add a comment here about what's going on though?  Just for future nerds wanting to make changes.  I stared at this for about 5min until I figured it out.  In my case I hadn't realised that you'd taken an actual filename (from line 486) and replaced it with a transformed filename (488).  Just some a line or two that explains what's going on would be handy.

src/documents/models.py
```diff
@@ -483,8 +483,14 @@ def from_path(cls, path):
...
+        for (pattern, repl) in settings.FILENAME_PARSE_TRANSFORMS:
+            (filename, count) = pattern.subn(repl, filename)
+            if count:
```
danielquinn: If I'm reading this right, if the for loop exits with `count` never being greater than `0`, then `filename` will just be whatever is last in the list of parse transforms.  It'd be safer to write it something like this:

```python
filename = os.path.basename(path)
transformed_filename = ""

for (pattern, repl) in settings.FILENAME_PARSE_TRANSFORMS:
    (transformed_filename, count) = pattern.subn(repl, filename)
    if count:
        break

for regex in cls.REGEXES.values():
    m = regex.match(transformed_filename or filename)
```

src/paperless/settings.py
```diff
@@ -317,6 +319,15 @@ def __get_boolean(key, default="NO"):
 DATE_ORDER = os.getenv("PAPERLESS_DATE_ORDER", "DMY")
 FILENAME_DATE_ORDER = os.getenv("PAPERLESS_FILENAME_DATE_ORDER")
 
+# Transformations applied before filename parsing
+FILENAME_PARSE_TRANSFORMS = []
+_filename_parse_transforms = os.getenv("PAPERLESS_FILENAME_PARSE_TRANSFORMS")
+if _filename_parse_transforms:
+    FILENAME_PARSE_TRANSFORMS = [(
```
danielquinn: Rather than a multi-line list comprehension, could you please just use a for loop?  Generally I want to go for readability over code you can fit into one statement.

grembo: I'll change it so it matches your style preferences.

src/documents/models.py
grembo: How does this sound to you as a comment?

> Apply at most one transformation configured in
> FILENAME_PARSE_TRANSFORMS to filename, before it is
> consumed by the parser. This allows to adapt custom
> filename formats to what paperless expects.
> (see [...])

src/documents/models.py
grembo: > If I'm reading this right, if the for loop exits with `count` never being greater than `0`, then `filename` will just be whatever is last in the list of parse transforms.

Nope, it won't. `re.subn` is like search/replace. It returns the modified input string, and the number of matches (= replace operations) that happened. If there are no matches, no modifications happened, count is zero and the returned string equals the string passed into the function. It never returns the transformation pattern (`repl`) it tried to apply.

Example at the python cli to demonstrate:

```
>>> import re
>>> re.compile('more').subn('less', 'papermore')
('paperless', 1)
>>> re.compile('less').subn('less', 'papermore')
('papermore', 0)
```

So there are two possible outcomes in each interation of the `for` loop:

  1. There was a match and `re.subn` altered the filename. In this case `count` is *not* zero and therefore we break out of the loop, as we only want to apply at most one transformation.

  2. There was no match, `filename` stays unaltered (as `re.subn` returned the unaltered string) and we go to the next iteration. If this happened to be the last iteration, the `for` loop ends naturally and `filename` still equals the value it had before entering the loop (which is `os.path.basename(path)`)

See also: https://docs.python.org/3/library/re.html#re.subn

grembo:
> I've asked for a couple changes. Some of them cosmetic, and one show-stopper. I'd also like to request that you add a test for `Document.from_path()` to account for your changes here.

@danielquinn I think I should've addressed all requested changes/concerns now.
@pitkley Sorry for piecemealing today's commits.

src/documents/models.py
danielquinn: Excellent.  Thank you for walking me through that.  As my mother likes to say, you learn something new every day :-)  This all looks good.  Sorry to complicate things.

danielquinn:
Thanks so much for this contribution!  Paperless has been getting some top-notch feature additions lately.

grembo:
You're welcome! Paperless is a really cool project :)

PR #545, opened 2019-06-01
Add insensitive search for documents

### LINKED ISSUE #115: Search is too rigid: allow accent-insensitive search queries
We are all used to "google-like" search, where accents and even typos in the query string do not really affect the search results. However Paperless seems to be a little bit too rigid when it comes to search. For instance, `Zürich` or `Telefónica` doesn't match `Zurich` and `Telefonica`, a behaviour which can frustrate some searches.

It looks like this can be achieved with the current SQLite backend by [providing some conversion tables](http://www.sqlite.org/faq.html#q18) that I guess they are so common that they already exist in the open source community. And after a quick look at Stack Overflow I [found](http://stackoverflow.com/questions/26871674/sqlite-custom-accent-collation-function-and-like-queries) [some](http://stackoverflow.com/questions/31331629/implementing-accent-insensitive-search-on-django-using-sqlite) [insight](http://stackoverflow.com/questions/2392732/sqlite-python-unicode-and-non-utf-data) which I hope they help.

CkuT:
This PR adds an insensitive-search on document title and content. This should close #115.

As it was pointed out by @danielquinn, `str.casefold()` is a good start, but not enough: it does not remove accents. For the latter, we need a `NFKD` unicode normalization (see https://unicode.org/reports/tr15/).

I'm open to any improvements !


CkuT:
I think the search by tag or correspondant is now broken if there are any accents. The PR is still in WIP.

src/documents/admin.py
```diff
@@ -13,12 +13,11 @@
 from django.utils.safestring import mark_safe
 from djangoql.admin import DjangoQLSearchMixin
 
-from documents.actions import (
-    add_tag_to_selected,
-    remove_correspondent_from_selected,
-    remove_tag_from_selected,
-    set_correspondent_on_selected
-)
+from documents.actions import (add_tag_to_selected,
```
danielquinn: In the interests of consistency, please don't reformat imports.  If anything, imports should conform to an isort configuration of:

```bash
isort -m 3 --dont-skip __init__.py --virtual-env ${virtualenv}/.. --quiet ${file_name}
```

src/documents/migrations/0023_document_searchable_content.py
```diff
@@ -0,0 +1,44 @@
+# Generated by Django 2.0.13 on 2019-05-30 14:50
+import unicodedata
+
+from django.db import migrations, models
+
+from paperless.utils import slugify as slugifyOCR
```
danielquinn: While it can be tempting to import stuff from modules into a migration to keep your code DRY, this will later bite us in the ass should we decide to rename/remove this function.  It will break all migrations for all time as a result.

Instead, if you have logic you wish to make available in a migration, copy it verbatim into the migration.

src/paperless/utils.py
```diff
@@ -0,0 +1,9 @@
+import unicodedata
+
+
+def slugify(content):
```
danielquinn: Can we come up with a more appropriate name for this than `slugify()`?  Something like `make_searchable()` would be more apt as this doesn't turn the content into a slug at all.

src/documents/migrations/0023_document_searchable_content.py
```diff
@@ -0,0 +1,44 @@
...
+    def casefold_forwards(apps, schema_editor):
+        Document = apps.get_model("documents", "Document")
+        for doc in Document.objects.all():
+            if doc.title is not None:
+                doc.searchable_title = slugifyOCR(doc.title)
+            if doc.content is not None:
+                doc.searchable_content = slugifyOCR(doc.content)
+            doc.save()
+
+    def casefold_backwards(apps, schema_editor):
```
danielquinn: Protip: you don't need to create an empty backwards method.  You can just do `migrations.RunPython(casefold_forwards, migrations.RunPython.noop)` below instead.

src/documents/tests/test_document_model.py
```diff
@@ -21,3 +21,28 @@ def test_file_deletion(self):
             mock_unlink.assert_any_call(file_path)
             mock_unlink.assert_any_call(thumb_path)
             self.assertEqual(mock_unlink.call_count, 2)
+
```
danielquinn: Hooray for tests!

src/documents/models.py
```diff
@@ -266,6 +281,13 @@ def __str__(self):
             return "{}: {}".format(created, self.correspondent or self.title)
         return str(created)
 
+    def save(self, *args, **kwargs):
```
danielquinn: Thank you for doing this in `.save()` and not in a signal.  This is much more appropriate.

CkuT:
Hey ! Does someone have any time to review this ? :)

src/paperless/utils.py
```diff
@@ -0,0 +1,9 @@
+import unicodedata
+
+
+def make_searchable(content):
+    return (
+        unicodedata.normalize("NFKD", content.casefold())
+        .encode("ASCII", "ignore")
```
pitkley: Hm, I'm not sure if this is the right thing to do. In light of [your change only searching the new searchable fields](https://github.com/the-paperless-project/paperless/pull/545/files#diff-8515db01a71e5ae1a53884301d7dc1f7L176-R182), this straight up breaks search for languages that share no characters with the ASCII codepage, right? Judging from the [official Python docs](https://docs.python.org/3/howto/unicode.html#converting-to-bytes), `"ignore"` would simply drop non-ASCII characters, resulting in a potentially empty search-term.

As far as I can tell not ignoring these characters here would break the entire feature, your `Zürich Weiß` example would, after case-folding first, produce `zu\xcc\x88rich weiss`, which obviously cannot be searched for with `zurich weiss`.

Maybe the simplest fix is to search both the new ASCII-fied fields _and_ the regular fields?

PR #552, opened 2019-07-24
Use whitenoise and brotli for static file compression & caching

diveflo:
[Whitenoise](http://whitenoise.evans.io/en/stable/index.html) can significantly improve the performance of the Django webserver by optimizing the static file serving. Additionally, this PR enables [brotli compression](https://github.com/google/brotli) for these files.

This requires the collectstatic command to be done even with the initial bare-metal setup. In the Dockerfile this is done automatically (thank you @Strubbl in @ #353).

I would afterwards open a PR to enable gunicorn also for the docker setup as it can then 'just work' even without nginx or Apache.

Please let me know if you have any issues, I couldn't get pipenv to add whitenoise and brotli with less impact on the remaining packages. Maybe somebody can have a look at this:) Thanks!


Strubbl:
Cool idea. If we merge this, we can close #353

ddddavidmartin:
Thanks for the pull request, this sounds great! I'm running your branch on my docker image and it seems to work fine.

I just have one question: [how come the dependencies in the `Pipfile` are now all versioned](https://github.com/the-paperless-project/paperless/pull/552/files#diff-1e61a31bf9b94805f869dc4137ec1885) when they weren't before? Does it matter?

diveflo:
That is one of the issues I had with pipenv that I couldn’t really figure out.
I can manually remove the versions from the Pipfile but I couldn’t get the environment to work locally like this, though this could be caused by the fringe-hardware (RockPi 4) that I’m using.
I’ll commit those changes tonight and maybe you could give it another try like that?
Thanks!

ddddavidmartin:
Hmmm, with your latest changes I'm getting an error when I build the docker image on my Mac. The commit with the explicit version numbers builds fine.

```
Successfully installed certifi-2019.6.16 pip-19.2.1 pipenv-2018.11.26 virtualenv-16.7.2 virtualenv-clone-0.5.3
Your Pipfile.lock (c3161e) is out of date. Expected: (336d00).
Traceback (most recent call last):
  File "/usr/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pipenv/vendor/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pipenv/cli/command.py", line 254, in install
    editable_packages=state.installstate.editables,
  File "/usr/lib/python3.6/site-packages/pipenv/core.py", line 1874, in do_install
    keep_outdated=keep_outdated
  File "/usr/lib/python3.6/site-packages/pipenv/core.py", line 1192, in do_init
    raise exceptions.DeployException
  File "/usr/lib/python3.6/site-packages/pipenv/exceptions.py", line 168, in __init__
    PipenvUsageError.__init__(message=fix_utf8(message), extra=extra, **kwargs)
TypeError: __init__() missing 1 required positional argument: 'self'
The command '/bin/sh -c apk add --no-cache       bash       curl       ghostscript       gnupg       imagemagick       libmagic       libpq       optipng       poppler       python3       shadow       sudo       tesseract-ocr       unpaper &&     apk add --no-cache --virtual .build-dependencies       g++       gcc       jpeg-dev       musl-dev       poppler-dev       postgresql-dev       python3-dev       zlib-dev &&     python3 -m ensurepip &&     rm -r /usr/lib/python*/ensurepip &&     cd /usr/src/paperless &&     pip3 install --upgrade pip pipenv &&     pipenv install --system --deploy &&     apk del .build-dependencies &&     mkdir -p $PAPERLESS_CONSUMPTION_DIR &&     addgroup -g 1000 paperless &&     adduser -D -u 1000 -G paperless -h /usr/src/paperless paperless &&     chown -Rh paperless:paperless /usr/src/paperless &&     mkdir -p $PAPERLESS_EXPORT_DIR &&     chmod 755 /sbin/docker-entrypoint.sh' returned a non-zero code: 1
```

I'm not really familiar with this though. So from my side we can leave the version numbers explicit.

diveflo:
Yeah sorry - that was my shitty attempt at fixing it manually.
I’ll try it again tonight, but pipenv won’t let me add new dependencies (whitenoise and brotli) without fixing the version numbers in the Pipfile

diveflo:
I fixed the version information in the Pipfile and reduced the number of dependency updates.
However, a few of the necessary updates are not compatible with Python 3.4.
For example shinx was already not compatible with it - even before the new version.
If the removal of support for Python 3.4 is okay I think we can go ahead here. I already removed Python 3.4 from the Travis build yml.

danielquinn:
Even Debian stable is on Python 3.5, so I think dropping 3.4 is safe.

diveflo:
@danielquinn sounds good:)
I removed one 'reference' to python3.4 in my last commit.
If you have the time I'd appreciate a review of the changes! I tested pretty thoroughly in my docker setup but I only checked the basics on a native install.
@ddddavidmartin as you've already tested it, could it give it another go and have a look through the code changes?
@Strubbl as part of the changes are your work I'd appreciate a quick check from you as well:)

Afterwards I'll put up the pull request to enable gunicorn on Docker.

Thanks!

ddddavidmartin:
Thanks for your efforts @diveflo! This looks really good now! I tried it on my Mac and on my Synologoy DSM through docker and it works fine.

I'd be really happy to see this merged in.

src/paperless/settings.py
```diff
@@ -101,6 +103,9 @@ def __get_boolean(key, default="NO"):
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 ]
 
+# Enable whitenoise compression and caching
+STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```
MasterofJOKers: Is that something I can use without django's runserver, i.e. when I serve static files via Nginx?

diveflo: My understanding is yes, because whitenoise also requires you to collect your static files once and then compresses them but leaves the “original” static files intact.
However, I haven’t tried it yet.
Would it be possible for you to test this on your system? 
Unfortunately I have not gotten Nginx & static files to work properly (-> why I opted for this approach 😅)

scripts/docker-entrypoint.sh
```diff
@@ -55,10 +55,15 @@ migrations() {
...
 
+collectstatic() {
+    sudo -HEu paperless "/usr/src/paperless/src/manage.py" "collectstatic" "--clear" "--no-input"
+}
+
 initialize() {
     map_uidgid
     set_permissions
     migrations
+    collectstatic
```
pitkley: Rather than collecting the static files on every startup, since they depend on the code-base that doesn't change dynamically, I'd suggest collecting them within the Dockerfile. This offloads the work from the user to the builder, while only adding ~10MB in image size (485MB before vs. 495MB after), which I think is worth it.

This is my suggested (and tested) change:

```diff
diff --git a/Dockerfile b/Dockerfile
index cfcfbd2..d5400aa 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,3 +65,6 @@ COPY src/ /usr/src/paperless/src/
 COPY data/ /usr/src/paperless/data/
 COPY media/ /usr/src/paperless/media/

+# Collect static files
+RUN sudo -HEu paperless /usr/src/paperless/src/manage.py collectstatic --clear --no-input
+
diff --git a/scripts/docker-entrypoint.sh b/scripts/docker-entrypoint.sh
index 303d1b8..5cfcd40 100644
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -55,15 +55,10 @@ migrations() {
     fi
 }

-collectstatic() {
-    sudo -HEu paperless "/usr/src/paperless/src/manage.py" "collectstatic" "--clear" "--no-input"
-}
-
 initialize() {
     map_uidgid
     set_permissions
     migrations
-    collectstatic
 }

 install_languages() {
```

diveflo: That's a great idea - thank you! I'll give it a whirl on my setup and integrate your suggested changes asap.

diveflo:
> Thanks for your efforts @diveflo! This looks really good now! I tried it on my Mac and on my Synologoy DSM through docker and it works fine.
> 
> I'd be really happy to see this merged in.

Hi @ddddavidmartin could you check the latest changes wrt to the move of static-file creation to the docker file and re-do your review? Thanks!

ddddavidmartin:
Nice thinking @pitkley and thanks for the rework @diveflo! I tested this again on my Macbook and Docker through Synology and it works fine. The startup of the docker container is a lot quicker now without the processing of all the files.

ddddavidmartin:
Hmmm, I don't think my review counts as I don't have write access to the repository.

@danielquinn, would it be worth relaxing the rule on required reviews to include reviews by read-only accounts? Incoming changes would still have to be merged by someone with write access so that seems okay to me. Thanks!

danielquinn:
I'm not comfortable relaxing the rules for read-only reviews because it could mean that just one maintainer could create another read-only account, approve a change from both accounts and merge it without sufficient oversight.  I much prefer the "everyone watches everyone" model ;-)

With that said though, @ddddavidmartin, I'm surprised I haven't given you write permissions!  I shall hook you up.

As for this branch, I'm going to merge it now.  Thanks so much for your contribution @diveflo!  This should help a lot with performance.

PR #556, opened 2019-08-11
Don't parse dates with more than 4 digits for the year

heinrich5991:
The regex was broken before, using `(?!…)` instead of `(?<=…)`.


MasterofJOKers:
Why do we need to lookbehind and lookahead? Can't we get away with something like this?
```python
r = re.compile(r'(?:\b|[_-])('\
               r'(?:[0-9]{1,2}[./-][0-9]{1,2}[./-](?:[0-9]{4}|[0-9]{2}))|'\
               r'(?:(?:[0-9]{4}|[0-9]{2})[./-][0-9]{1,2}[./-][0-9]{1,2})|'\
               r'(?:[0-9]{1,2}\. +[^\W\d_]{3,9} (?:[0-9]{4}|[0-9]{2}))|'\
               r'(?:[^\W\d_]{3,9}(?: [0-9]{1,2},)? [0-9]{4})'\
               r')(?:\b|[_-])')
```
In some manual testing, it seems to match everything matched in the unit tests. We can then use `m.group(1)` for the date-part of the matched string.

:+1: for the additional tests.

heinrich5991:
Updated with the suggestion to not use lookahead/lookbehind.

src/documents/parsers.py
```diff
@@ -21,11 +21,13 @@
...
-    r'(\b|(?!=([_-])))([0-9]{1,2})[\.\/-]([0-9]{1,2})[\.\/-]([0-9]{4}|[0-9]{2})(\b|(?=([_-])))|' +  # NOQA: E501
-    r'(\b|(?!=([_-])))([0-9]{4}|[0-9]{2})[\.\/-]([0-9]{1,2})[\.\/-]([0-9]{1,2})(\b|(?=([_-])))|' +  # NOQA: E501
-    r'(\b|(?!=([_-])))([0-9]{1,2}[\. ]+[^ ]{3,9} ([0-9]{4}|[0-9]{2}))(\b|(?=([_-])))|' +  # NOQA: E501
-    r'(\b|(?!=([_-])))([^\W\d_]{3,9} [0-9]{1,2}, ([0-9]{4}))(\b|(?=([_-])))|' +
-    r'(\b|(?!=([_-])))([^\W\d_]{3,9} [0-9]{4})(\b|(?=([_-])))'
+    r'(?:\b|[_-])(' +
+    r'(?:[0-9]{1,2})[\.\/-](?:[0-9]{1,2})[\.\/-](?:[0-9]{4}|[0-9]{2})|' +
+    r'(?:[0-9]{4}|[0-9]{2})[\.\/-](?:[0-9]{1,2})[\.\/-](?:[0-9]{1,2})|' +
+    r'(?:[0-9]{1,2}[\. ]+[^ ]{3,9} (?:[0-9]{4}|[0-9]{2}))|' +
+    r'(?:[^\W\d_]{3,9} [0-9]{1,2}, (?:[0-9]{4}))|' +
```
MasterofJOKers: The `(?:)` at the year here are superfluous.

heinrich5991:
Removed all the superfluous `(?:)`.

MasterofJOKers:
> Removed all the superfluous `(?:)`.

Great, could you also remove the superfluous `\` in the `[]`, while you're at it?

heinrich5991:
Removed the superfluous backslashes in the regex.

PR #559, opened 2019-08-29
Serve thumbnails additionally in WebP format to improve performance

diveflo:
Besides django itself, the paperless webserver is often a bit sluggish due to the large thumbnails, oftentimes summing up to more than 20 MB per page.
Optipng is already used to reduce this network load a bit but uncompressed pngs are still very large and not state-of-the-art.
Google's [WebP](https://developers.google.com/speed/webp/) is a more modern image format supported by most* browsers. The big exception is Apple's Safari.
Hence in this pull-request I suggest we offer in the meantime WebP and png. Using the `<picture>` tag instead of `` [the webserver can offer both with a default](https://github.com/radek-anuszewski/HTML-picture-tag-in-practice/blob/master/index.html) to the most-compatible option (here: png). In my initial experience the file size of the thumbnails is around four times smaller compared to png.

**Hints/Limitations**
- no migration implemented (yet?) -> this will not generate webp thumbnails for your existing documents. However, it should not break anything. You'll just have webp & png for your new documents and pngs for your old ones.
- not 100% happy with the quite enlarged Documents API. I'm thinking about refactoring the thumbnail logic out of the document and parser as these should be separate concerns. This could also be the chance to make this a bit more generic and possibly support a multitude of formats in the future. (E.g. Apple will probably not support WebP anytime soon as they have put their money on HEIC)

Let me know what you think! Comments (and reviews!) welcome:)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.10
 
 LABEL maintainer="The Paperless Project https://github.com/the-paperless-project/paperless" \
       contributors="Guy Addadi <addadi@gmail.com>, Pit Kleyersburg <pitkley@googlemail.com>, \
@@ -21,6 +21,7 @@ RUN apk add --no-cache \
       libmagic \
       libpq \
       optipng \
+      libwebp \
       poppler \
       python3 \
       shadow \
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -284,6 +284,7 @@ class DocumentAdmin(DjangoQLSearchMixin, CommonAdmin):
         html_thumb_srcset = self._html_tag(
                     "source",
                     srcset = reverse("fetch", kwargs={"kind": "thumb", "pk": obj.pk}),
+                    type = "image/png",
                     width=180,
                     alt="Thumbnail of {}".format(obj.file_name),
                     title=obj.file_name
@@ -292,6 +293,7 @@ class DocumentAdmin(DjangoQLSearchMixin, CommonAdmin):
         html_thumbwebp_srcset = self._html_tag(
                     "source",
                     srcset= reverse("fetch", kwargs={"kind": "thumbwebp", "pk": obj.pk}),
+                    type = "image/webp",
                     width=180,
                     alt="Thumbnail of {}".format(obj.file_name),
                     title=obj.file_name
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -76,6 +76,7 @@ db.sqlite3-journal
 virtualenv
 docker-compose.yml
 docker-compose.env
+.vscode/
```
danielquinn: As this isn't project-specific, I'd suggest that you add `.vscode` to your personal git ignore file (on Linux systems, this is at `${HOME}/.config/git/ignore`).

diveflo: Oh, I totally missed that - thanks! Rebase is on the way.

diveflo:
> This all looks really good, and I really appreciate your thoroughness with including the import/export management commands along with everything else.
> 
> I have a few reservations with how we're referencing everything as `thumbnail` and `thumbnail_webp` all over the place, as that feels like a first step toward a sprawling set of special cases. I think I would have preferred to go with just one of webp, png, or jpg. However as you mentioned, [webp is still not sufficiently adopted](https://caniuse.com/#search=webp) to go all-in so this might be the best option.
> 
> Anyway, aside from the minor change request in the .gitignore file and the rejections from Travis on account of the pep8 violations (please conform to the [style guide](https://paperless.readthedocs.io/en/latest/contributing.html#pep8)) I don't have any real objections here. Anyone else have thoughts?

Yeah, I totally agree about the code cluttering. Alternatively, we could just go for jpg as an interim which would at least reduce the size about ~50% and I'll try to come up with some cleaner abstraction for the thumbnails which internally handles different/multiple formats. My daily bread & butter engineering is unfortunately C++/CLI/C# rather than Python so that might take a bit.
If you have a clever idea just let me know!

danielquinn:
If you think we can get 50% of out a switch to JPEG, that might be the cleanest & best option rather than this two-format option using webp & png.  If we went that route, all old thumbnails would remain in png format, but new ones would be stored as jpgs, and we could add a script to re-generate all thumbnails for people chasing performance.

diveflo:
I'll run it on my (modest) document collection and let you know what the actual numbers are.

diveflo:
@danielquinn I had a look at 10 examples and the jpegs are between 15 and 50% of the size of the (optimised) pngs. I'll close this PR until I've figured out a more elegant multi-format solution and open a new one for the jpeg route! Thanks for the feedback!

danielquinn:
Cool beans, and thanks for your efforts!

PR #564, opened 2019-09-08
Gunicorn webserver in Docker

diveflo:
In #552 we prepared django to serve the static files in a more efficient manner using whitenoise.
Next step: using [gunicorn](https://gunicorn.org/) to spin up a few more workers and get that webserver flying:)

I used the existing gunicorn config file, however I had to change the bind-ip to 0.0.0.0 to get it working. Does anybody know if that's really necessary?

Reviews, comments etc. are always welcome!

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -99,7 +99,12 @@ if [[ "$1" != "/"* ]]; then
         install_languages "$PAPERLESS_OCR_LANGUAGES"
     fi
 
-    exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
+    if [[ "$1" = "gunicorn" ]]; then
+        cd /usr/src/paperless/src/ && \
+            exec sudo -HEu paperless /usr/bin/gunicorn -c gunicorn.conf paperless.wsgi
+    else
+        exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
+    fi
 fi
 
 exec "$@"
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -100,8 +100,9 @@ if [[ "$1" != "/"* ]]; then
     fi
 
     if [[ "$1" = "gunicorn" ]]; then
+        shift
         cd /usr/src/paperless/src/ && \
-            exec sudo -HEu paperless /usr/bin/gunicorn -c /usr/src/paperless/gunicorn.conf paperless.wsgi
+            exec sudo -HEu paperless /usr/bin/gunicorn -c /usr/src/paperless/gunicorn.conf "$@" paperless.wsgi
     else
         exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
     fi
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -101,7 +101,7 @@ if [[ "$1" != "/"* ]]; then
 
     if [[ "$1" = "gunicorn" ]]; then
         cd /usr/src/paperless/src/ && \
-            exec sudo -HEu paperless /usr/bin/gunicorn -c gunicorn.conf paperless.wsgi
+            exec sudo -HEu paperless /usr/bin/gunicorn -c /usr/src/paperless/gunicorn.conf paperless.wsgi
     else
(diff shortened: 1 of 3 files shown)


danielquinn:
I don't think the default of `127.0.0.1` will work as communication with a Docker container is usually done over the network Docker establishes between containers.  The problem is that you don't typically have that IP before the container is started, so hard-coding an IP into the config isn't doable, hence the use of `0.0.0.0` instead.  However, since containers aren't directly accessible except by way of setting a port from a `docker-compose.yml` file, I don't see any downside to using `0.0.0.0` in this case.  @pitkley is the authority on Docker stuff in this project though, so I bow to his expertise in this area ;-)

ddddavidmartin:
This is very exciting, thanks for bringing this in @diveflo!

How can I tell whether it is working/running? The only difference I can see is that at startup IP 0.0.0.0 is used.

Screen Shot 2019-09-11 at 7 00 00 pm

Thanks!

diveflo:
During the docket startup it should say guncorn starting and booting 4 workers with pid xyz 
Your log actually looks like the old django runserver.
Did you rebuild the docker image?
‘docker-compose up -d --build’

ddddavidmartin:
Thanks for clarifying! My mistake was that I set up the container from a backup after updating the image and it was still using the old `"runserver", "--insecure", "--noreload", "0.0.0.0:8000"` command. It is working fine now.

scripts/gunicorn.conf
```diff
@@ -1,6 +1,6 @@
-bind = '127.0.0.1:8000'
+bind = '0.0.0.0:8000'
 backlog = 2048
-workers = 1
+workers = 4
```
ddddavidmartin: Out of curiosity: Are the hardcoded 4 workers something I should change depending on my architecture or is it a sane default?

diveflo: I took it as an “okay” default for most systems.
This [blog post](https://medium.com/building-the-system/gunicorn-3-means-of-concurrency-efbb547674b7) gives a rule of thumb of 2xCPUs + 1.
They also talk about different worker types for different workloads. I wanted to look into this next:)

ddddavidmartin: [The official documentation seems pretty good as well and has some more details on the workers and everything](http://docs.gunicorn.org/en/latest/design.html#how-many-workers). 

I guess for Paperless there is often just one client connecting to it. So I would probably be more concerned about having too many workers for a simple system than not enough.

ddddavidmartin:
Nice work @diveflo!

It's great to get a more proper webserver going with Paperless! The changes to the code are minimal, the original Django server still works with the original command and it works without issues here on my docker setup.

scripts/gunicorn.conf
```diff
@@ -1,6 +1,6 @@
-bind = '127.0.0.1:8000'
+bind = '0.0.0.0:8000'
```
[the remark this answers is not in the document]
diveflo: Thanks! I’ll give it a try this weekend

scripts/docker-entrypoint.sh
```diff
@@ -99,7 +99,12 @@ if [[ "$1" != "/"* ]]; then
         install_languages "$PAPERLESS_OCR_LANGUAGES"
     fi
 
-    exec sudo -HEu paperless "/usr/src/paperless/src/manage.py" "$@"
+    if [[ "$1" = "gunicorn" ]]; then
+        cd /usr/src/paperless/src/ && \
+            exec sudo -HEu paperless /usr/bin/gunicorn -c /usr/src/paperless/gunicorn.conf paperless.wsgi
```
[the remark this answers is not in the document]
diveflo: Hi! I pushed some new changes, which basically do what you suggested here. But the bind ip now comes through the command field in the docker-compose.yml, similar to how it is done by the django dev-server, so we have some consistency here.

scripts/gunicorn.conf
diveflo: I removed the worker increase from this PR, separation of concerns etc. For most systems this will still make sense, so maybe we should open a separate PR for this.

scripts/gunicorn.conf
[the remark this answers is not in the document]
diveflo: See discussion below. Gunicorn.conf is now back to `127.0.0.1` for bare-metal install and the docker-compose.yml injects `0.0.0.0` via the command field.

Issue #576, opened 2019-10-26
Support for SSL


fdw:
When using the docker-compose file, I can't find any setting to enable SSL, neither in the readme.
Is this already supported?

If not, I would like to request this feature.

ddddavidmartin:
AFAIK there is no support for SSL. From what I have seen people usually don't expose Paperless to the internet. I have all my personal documents and scans in there and I don't trust myself or this project to keep it safe. I use an SSH tunnel to access Paperless from outside my network.

lbischof:
Or use a reverse proxy like nginx or traefik

fdw:
I don't have it connected to the internet, either. However, all other internal services I use do support SSL natively without the need for a reverse proxy. Since Paperless also has more private data, I think it makes sense to offer this functionality as well.

MasterofJOKers:
It looks like we could support this in Paperless' docker container, if we wanted to, as [gunicorn supports ssl](https://stackoverflow.com/questions/7406805/running-gunicorn-on-https). We would have to generate some cert and key, though.

fdw:
@MasterofJOKers Sounds fine to me :) I don't even think that the cert/key should be generated, but just a configuration for the user.

MasterofJOKers:
I haven't tested this, but somethin along the lines of this should work. You would have to set `PAPERLESS_USE_SSL=true` in the docker-compose.env, I guess. Never used docker-compose.
I've hardcoded the paths to the cert/key file here, because we somehow need to get them into the container anyways.
```diff
diff --git a/scripts/docker-entrypoint.sh b/scripts/docker-entrypoint.sh
index 14f431b..b09b420 100644
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -101,8 +101,14 @@ if [[ "$1" != "/"* ]]; then
 
     if [[ "$1" = "gunicorn" ]]; then
         shift
+        EXTRA_PARAMS=""
+        SSL_KEY_PATH="/usr/src/paperless/data/ssl.key"
+        SSL_CERT_PATH="/usr/src/paperless/data/ssl.cert"
+        if [ "${PAPERLESS_USE_SSL}" = "true" ] && [ -f "${SSL_KEY_PATH}" ] && [ -f "${SSL_CERT_PATH}" ]; then
+            EXTRA_PARAMS="--certfile='${SSL_CERT_PATH}' --keyfile='${SSL_KEY_PATH}'"
+        fi
[...]

PR #579, opened 2019-11-03
Support SSL for web interface

fdw:
Thanks again to @MasterofJOKers for the actual solution (#576).

Is this sufficient for the documentation?

### DIFF
--- a/docker-compose.env.example
+++ b/docker-compose.env.example
@@ -16,6 +16,9 @@
 # different from PAPERLESS_OCR_LANGUAGE (default=eng), which defines the
 # default language used when guessing the language from the OCR output.
 # PAPERLESS_OCR_LANGUAGES=deu ita
+
+# Set Paperless to use SSL for the web interface.
+# Enabling this will require ssl.key and ssl.cert files in paperless' data directory.
 # PAPERLESS_USE_SSL=false
 
 # You can change the default user and group id to a custom one
--- a/docker-compose.env.example
+++ b/docker-compose.env.example
@@ -16,6 +16,7 @@
 # different from PAPERLESS_OCR_LANGUAGE (default=eng), which defines the
 # default language used when guessing the language from the OCR output.
 # PAPERLESS_OCR_LANGUAGES=deu ita
+# PAPERLESS_USE_SSL=false
 
 # You can change the default user and group id to a custom one
 # USERMAP_UID=1000
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -193,6 +193,11 @@ Docker Method
      container and thus the one of the consumption directory. Furthermore, you
      can change the id of the default user as well using ``USERMAP_UID``.
 
+  ``USE_SSL``
+    If you want Paperless to use SSL for the user interface, set this variable
+    to ``true``. You also need to copy your certificate and key to the ``data``
+    directory, named ``ssl.cert`` and ``ssl.key``.
+
 6. Run ``docker-compose up -d``. This will create and start the necessary
    containers.
 7. To be able to login, you will need a super user. To create it, execute the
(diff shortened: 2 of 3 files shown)


docs/setup.rst
```diff
@@ -193,6 +193,11 @@ Docker Method
      container and thus the one of the consumption directory. Furthermore, you
      can change the id of the default user as well using ``USERMAP_UID``.
 
+  ``USE_SSL``
```
MasterofJOKers: This should be `PAPERLESS_USE_SSL`, shouldn't it?

docker-compose.env.example
```diff
@@ -16,6 +16,7 @@
 # different from PAPERLESS_OCR_LANGUAGE (default=eng), which defines the
 # default language used when guessing the language from the OCR output.
 # PAPERLESS_OCR_LANGUAGES=deu ita
+# PAPERLESS_USE_SSL=false
```
MasterofJOKers: Please add some comment above this line, too, like
```
Enabling this will require ssl.key and ssl.cert files in paperless' data directory.
```

scripts/docker-entrypoint.sh
```diff
@@ -101,8 +101,14 @@ if [[ "$1" != "/"* ]]; then
 
     if [[ "$1" = "gunicorn" ]]; then
         shift
+        EXTRA_PARAMS=""
+        SSL_KEY_PATH="/usr/src/paperless/data/ssl.key"
+        SSL_CERT_PATH="/usr/src/paperless/data/ssl.cert"
+        if [ "${PAPERLESS_USE_SSL}" = "true" ] && [ -f "${SSL_KEY_PATH}" ] && [ -f "${SSL_CERT_PATH}" ]; then
+            EXTRA_PARAMS="--certfile=${SSL_CERT_PATH} --keyfile=${SSL_KEY_PATH}"
+        fi
```
MasterofJOKers: We might wanna change this to print some error message, if the files are not found.
```sh
if [ "${PAPERLESS_USE_SSL}" = "true" ]; then
    if [ -f "${SSL_KEY_PATH}" ] && [ -f "${SSL_CERT_PATH}" ]; then
        EXTRA_PARAMS="--certfile=${SSL_CERT_PATH} --keyfile=${SSL_KEY_PATH}"
    else
        echo "Error: Could not find certfile in ${SSL_CERT_PATH} or keyfile in ${SSL_KEY_PATH}, but \$PAPERLESS_USE_SSL is true. Starting without SSL enabled.
    fi
fi
```

docs/setup.rst
```diff
@@ -193,6 +193,11 @@ Docker Method
...
+    If you want Paperless to use SSL for the user interface, set this variable
+    to ``true``. You also need to copy your certificate and key to the ``data``
+    directory, named ``ssl.cert`` and ``ssl.key``.
```
MasterofJOKers: Given the discussion in the ticket, we should maybe mention that this setup is not ideal and one should rather use e.g. Nginx for proper setups.

docs/setup.rst
fdw: Of course.

docs/setup.rst
fdw: I'm not really sure what to write here. If you (or anyone else) have any better ideas, please tell me :)

scripts/docker-entrypoint.sh
fdw: I have kept the definitions of `$SSL_KEY_PATH` and `$SSL_CERT_PATH`, as it makes the setup easier. Or should those be configurable, as well?

docs/setup.rst
MasterofJOKers: > If you want Paperless to use SSL directly - which is a discouraged setup, try using e.g. Nginx as reverse proxy terminating SSL instead - set this variable […]
Something like that maybe?

fdw: Have you seen the added sentence in my latest commit? I think it is similar enough.

MasterofJOKers: You're right, I didn't see that :)

scripts/docker-entrypoint.sh
MasterofJOKers: I'd leave that as is until someone comes along and needs it changed.

fdw:
Can I do something to get this merged?

fdw:
Thank you, @ahyear, and of course many thanks to @MasterofJOKers!

PR #581, opened 2019-11-19
added dependency on network.target

nxnjz:


pitkley:
@nxnjz what is the reason for this change? If it is that an internet connection is available, would we want to use `network-online.target` instead?

nxnjz:
@pitkley Mainly so that paperless-webserver can cleanly shutdown before the network goes down (during reboots, etc.)

`network-online.target` is not required when accessing paperless on localhost, so having a dependency on it would a bit excessive I think.

[man systemd.special](https://www.freedesktop.org/software/systemd/man/systemd.special.html#network.target)

Issue #293, opened 2018-01-30
Give stores documents a structured filename


BastianPoe:
Hi!

Use case: I have paperless running on my Synology NAS and want to be able to find documents stored in paperless via the universal search of the system.

This requires them to be stored unencrypted (see https://github.com/danielquinn/paperless/pull/254) and they ideally also have a “speaking” filename (such as “correspondent - date - tags”).

Any thoughts about this?

Bastian

phha:
Hi,

I'm currently working around this issue by running a cron job, that

1. Runs manage.py document_exporter
2. Runs https://github.com/phha/paperless-export-organizer which will create symlinks to the exported files with readable filenames and a meaningful folder hierarchy.

I suppose you could also run this as a post-consume hook instead of a cron job.

BastianPoe:
The idea of having symlinks actually sounds like a good idea. I will have to check, if this works for my use case.

I run paperless in a docker on a Synology NAS. The "universal search" allows me to search through the whole system and of course I would like search through all my documents. #254 allows to store unencrypted documents (good!) and now I would like to have the filename reflect, what we know about the document.

I will give this a try and report back.

PR #593, opened 2019-12-27
Give stored documents a structured and configurable filename

BastianPoe:
This feature (originally mentioned in #293) allows to configure directory and filename formats for files in storage. This is helpful, when accessing storage with means other than paperless.


BastianPoe:
@ovv @pitkley What can I do to help in the review process?

src/documents/models.py
```diff
@@ -314,6 +397,83 @@ def thumbnail_file(self):
...
+    def set_filename(self, filename):
+        if os.path.isfile(self.filename_to_path(filename)):
+            self.filename = filename
+
+
+def delete_empty_directory(directory):
+    if len(os.listdir(directory)) == 0:
+        try:
+            os.rmdir(directory)
+        except os.error:
```
pitkley: I appreciate the TOCTOU check, but in general in Python it is more common to ask for forgiveness than for permission. I suggest to remove the `len()` check altogether and simply try to remove the directory.

src/documents/models.py
```diff
@@ -314,6 +397,83 @@ def thumbnail_file(self):
...
+@receiver(models.signals.post_save, sender=Document)
+def update_filename(sender, instance, **kwargs):
+    if instance.filename is None:
+        return
+
+    # Build the new filename
+    new_filename = instance.source_filename_new()
+
+    # If the filename is the same, then nothing needs to be done
+    if instance.filename is None or \
```
pitkley: The `instance.filename is None` check is duplicated, you already checked it above.

src/documents/models.py
```diff
@@ -314,6 +397,83 @@ def thumbnail_file(self):
...
+            os.rmdir(directory)
+        except os.error:
+            # Directory not empty
+            pass
+
+
+@receiver(models.signals.m2m_changed, sender=Document.tags.through)
+@receiver(models.signals.post_save, sender=Document)
+def update_filename(sender, instance, **kwargs):
+    if instance.filename is None:
```
pitkley: Why do we not want to set the filename if it is not populated yet?

src/documents/models.py
```diff
@@ -314,6 +397,83 @@ def thumbnail_file(self):
...
+    # Build the new filename
+    new_filename = instance.source_filename_new()
+
+    # If the filename is the same, then nothing needs to be done
+    if instance.filename is None or \
+       instance.filename == new_filename:
+        return
+
+    # Check if filename needs changing
+    if new_filename != instance.filename:
```
pitkley: Again, this check is a duplicate of the check directly above and is thus unnecessary.

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+            # Directory existed already, ignore
+            pass
+
+        return filename
+
+    @property
+    def source_path(self):
+        return self.filename_to_path(self.source_filename)
 
+    def filename_to_path(self, filename):
```
pitkley: Since this doesn't need any class-association, I'd suggest making it a static method. (The call-sites have to be adjusted accordingly.)

```suggestion
    @staticmethod
    def filename_to_path(filename):
```

src/documents/models.py
```diff
@@ -314,6 +397,83 @@ def thumbnail_file(self):
     def thumbnail_url(self):
         return reverse("fetch", kwargs={"kind": "thumb", "pk": self.pk})
 
+    def set_filename(self, filename):
+        if os.path.isfile(self.filename_to_path(filename)):
+            self.filename = filename
+
+
+def delete_empty_directory(directory):
```
pitkley: This only deletes the most nested directory, which means intermediate directories a user might have specified will not be cleaned up. I'm not sure what the easiest fix for this would be while ensuring we don't delete any of the applications directories we might still need (even if they turn out to be empty, like the storage-root), but I think a user would expect this behaviour.

src/documents/models.py
```diff
@@ -314,6 +397,83 @@ def thumbnail_file(self):
...
+@receiver(models.signals.post_delete, sender=Document)
+def delete_files(sender, instance, **kwargs):
+    if instance.filename is None:
+        return
+
+    # Remove the document
+    old_file = instance.filename_to_path(instance.filename)
+
+    if os.path.isfile(old_file):
+        os.remove(old_file)
```
pitkley: There is a potential TOCTOU issue here: the file could be gone between when we check and when we remove. I'd instead suggest to `try` to call `os.remove()` and catch a `FileNotFoundError`, printing the warning message. If any other exception happens, we probably want to propagate it (as this code is doing already).

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+
+            if delimeter is -1:
+                continue
+
+            key = t.name[:delimeter]
+            value = t.name[delimeter+1:]
+
+            mydictionary[key] = value
+
+        return mydictionary
```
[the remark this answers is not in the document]
BastianPoe: Indeed, this was not properly documented. Basically, we have two ways to use tags in filenames:
* When your tags have the format KEY-VALUE or KEY_VALUE, then tag[KEY] selects this particular tag
* Otherwise, you can still select tags by index by using tags[INDEX]

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+
+        # Append .gpg for encrypted files
         if self.storage_type == self.STORAGE_TYPE_GPG:
-            file_name += ".gpg"
+            filename += ".gpg"
+
+        # Create directory for target
+        create_dir = self.filename_to_path(slugify(directory))
+        try:
+            os.makedirs(create_dir)
```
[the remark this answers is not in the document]
BastianPoe: done

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+                continue
+
+            key = t.name[:delimeter]
+            value = t.name[delimeter+1:]
+
+            mydictionary[key] = value
+
+        return mydictionary
+
+    def source_filename_new(self):
```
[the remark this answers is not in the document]
BastianPoe: done

src/documents/models.py
BastianPoe: No, because the file might not have been saved yet (during consumption) and we will fail otherwise. Added documentation.

src/documents/models.py
BastianPoe: done

src/documents/models.py
BastianPoe: done

src/documents/models.py
BastianPoe: It will no go all the way up to the originals directory and delete all empty directories along the way

src/documents/models.py
BastianPoe: done

src/documents/models.py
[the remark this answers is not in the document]
BastianPoe: done

src/documents/models.py
BastianPoe: done

src/documents/models.py
BastianPoe: done

BastianPoe:
@pitkley Thank you very much for the review. I have fixed all of the comments, simplified the code and added more tests. That was definitely worth it - thanks for the feedback!

src/documents/models.py
```diff
@@ -398,13 +398,18 @@ def set_filename(self, filename):
...
         try:
             os.rmdir(directory)
         except os.error:
-            # Directory not empty
-            pass
+            # Directory not empty, no need to go further up
+            return
+
+        # Cut off actual directory and go one level up
+        directory, tmp = os.path.split(directory)
```
pitkley: I think it is more established to name an unused/ignored value a single underscore, but this is a very minor nit-pick. :slightly_smiling_face: 

```suggestion
        directory, _ = os.path.split(directory)
```

src/documents/tests/test_file_handling.py
```diff
@@ -145,6 +145,37 @@ def test_directory_not_empty(self):
...
+
+        document.delete()
+
+        self.assertEqual(os.path.isfile(settings.MEDIA_ROOT +
+                         "/documents/originals/none/none/none-0000001.pdf"), False)
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals/none/none"), False)
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals/none"), False)
+
```
pitkley: Can you add an assert that verifies that `/documents/original` still exists? Then we would have the "accidental deletion" I was worried about fully covered! :+1:

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+    def generate_source_filename(self):
+        # Create filename based on configured format
+        if settings.PAPERLESS_FILENAME_FORMAT is not None:
+            path = settings.PAPERLESS_FILENAME_FORMAT.format(
+                   correspondent=slugify(self.correspondent),
+                   title=slugify(self.title),
+                   created=slugify(self.created),
+                   added=slugify(self.added),
+                   tag=defaultdict(str, self.many_to_dictionary(self.tags)),
+                   tags=self.many_to_list(self.tags))
```
pitkley: Index-out-of-bounds errors are not handled here, right? This means that if a scanned document does not have any tags applied (or simply not as many as the user references in their format-string) this will throw an exception, breaking document consumption.

This should probably be handled differently, although I'm not yet sure what good behaviour would be here.

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
-        file_name = "{:07}.{}".format(self.pk, self.file_type)
+    def generate_source_filename(self):
+        # Create filename based on configured format
+        if settings.PAPERLESS_FILENAME_FORMAT is not None:
+            path = settings.PAPERLESS_FILENAME_FORMAT.format(
+                   correspondent=slugify(self.correspondent),
+                   title=slugify(self.title),
+                   created=slugify(self.created),
+                   added=slugify(self.added),
+                   tag=defaultdict(str, self.many_to_dictionary(self.tags)),
```
pitkley: While this doesn't suffer from any `KeyError`s from occuring, thanks to the default dict, it does leave weird filenames if the format specified unconditional content, e.g.:

```
PAPERLESS_FILENAME_FORMAT={correspondent}-{tag[type]}
```

Would result in a file named e.g. `MyBank--0000000.pdf` -- notice the double-dashes. Even worse: if the user specified the filename to be something like `{tag[foo]}-{tag[bar]}` it could cause filenames like `--0000000.pdf` or `-bar-0000000.pdf`.

Now while leading dashes should work on most filesystems, at least handling them with a common shell is weird because [they might be interpreted as options](https://unix.stackexchange.com/questions/1519/how-do-i-delete-a-file-whose-name-begins-with-hyphen-a-k-a-dash-or-minus).

Considering my other comment about the next line potentially causing `IndexError`s, maybe the behaviour here should be different: try to format the string as provided, catching `KeyError`s and `IndexError`s. If either occurs, fallback to the default numbers-only filename.

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+    def source_filename(self):
+        if self.filename is None:
+            self.filename = self.generate_source_filename()
+
+        return self.filename
+
+    @staticmethod
+    def many_to_dictionary(field):
+        # Converts ManyToManyField to dictionary by assuming, that field
+        # entries contain an _ or - which will be used as a delimeter
```
pitkley: Just a minor nit: it is "delimiter", not "delimeter". Feel free to correct it if you want to, but it's no big deal. :slightly_smiling_face:

src/documents/models.py
```diff
@@ -267,17 +277,90 @@ def __str__(self):
...
+
+        return mydictionary
+
+    @staticmethod
+    def many_to_list(field):
+        # Converts ManyToManyField to list
+        mylist = list()
+
+        for t in field.all():
+            mylist.append(slugify(t))
```
pitkley: Does `slugify(t)` do the right thing, or should this be `slugify(t.name)`? I.e. does `str(t)` return the name of the tag? (Maybe we should be explicit here rather than rely on implicit behaviour that might change.)

src/documents/models.py
BastianPoe: I have added handling - admittedly in a rather clumsy way but workable and tested through unit tests. What do you think?

src/documents/models.py
BastianPoe: Good point - fixed by defaulting to "none" which ensure never to have a - as prefix. Also, -- will be prevented unless explicitly configured by the user.

src/documents/models.py
BastianPoe: Yep, thanks for the feedback - fixed.

src/documents/models.py
BastianPoe: Fixed (although the result for the moment is obviously the same, but being more explicit might prevent problems in the future).

BastianPoe:
@pitkley Again, thank you very much for taking the time to review and providing comments. This again increased the quality and further lead to me adding more tests (which is good!). I have addressed all your findings.

Could you please have another look and provide feedback or merge if all is ok?

@danielquinn @ovv @ahyear @MasterofJOKers @jat255 @bmsleight @CkuT: Fell free to also provide comments, I am happy to fix them.

src/documents/models.py
```diff
@@ -267,17 +277,111 @@ def __str__(self):
...
+
+    def generate_source_filename(self):
+        # Create filename based on configured format
+        if settings.PAPERLESS_FILENAME_FORMAT is not None:
+            tag = defaultdict(lambda: slugify(None),
+                              self.many_to_dictionary(self.tags))
+            list_length = 10
+            tags = self.many_to_list(self.tags)
+            while True:
+                tags = Document.fill_list(tags, list_length, None)
```
pitkley: Hm, I don't think I like this solution. The while-loops both here and in `fill_list` are pretty confusing if you read this code without prior knowledge.

A much simpler solution would be to "mis-use" a defaultdict and get this down to one line:

```python
tags = defaultdict(lambda: slugify(None),
                   enumerate(self.many_to_list(self.tags)))
```

This will create a dict indexed using the original list's indices. If an invalid index is provided, the default will be returned (`"none"`). (It does allow retrieval using non-integer indices, but the user shouldn't be doing that. Anyway, it will not cause errors, just return `"none"` as well.)

Doing it like that removes `fill_list` again and gets rid of the while-loop. In addition, the try-except can be removed as well.

pitkley: As an additional idea: how about the distinction between `tag` and `tags` is removed altogether, and we include both string and int indices in the defaultdict? I thought it could work like this:

```python
@staticmethod
def many_to_dictionary(field):
    # Converts ManyToManyField to dictionary by assuming, that field
    # entries contain an _ or - which will be used as a delimiter
    mydictionary = dict()

    for index, t in enumerate(field.all()):
        # Populate tag names by index
        mydictionary[index] = slugify(t.name)

        # Find delimiter
        delimiter = t.name.find('_')

        if delimiter is -1:
            delimiter = t.name.find('-')

        if delimiter is -1:
            continue

        key = t.name[:delimiter]
        value = t.name[delimiter+1:]

        mydictionary[slugify(key)] = slugify(value)

    return mydictionary

def generate_source_filename(self):
    if settings.PAPERLESS_FILENAME_FORMAT is not None:
        tags = defaultdict(lambda: slugify(None),
                           self.many_to_dictionary(self.tags))
        # ...
```

Then it could be used like this:

```
PAPERLESS_FILENAME_FORMAT="tags[foo]-tags[3]"
```

If I have not missed anything, I think this would give the same result. Let me know what you think! :slightly_smiling_face:

src/documents/tests/test_file_handling.py
```diff
@@ -0,0 +1,333 @@
...
+        self.assertEqual(os.path.isfile(settings.MEDIA_ROOT +
+                         "/documents/originals/none/none/none-0000001.pdf"),
+                         False)
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals/none/none"), False)
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals/none"), False)
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals"), False)
```
pitkley: Hm, shouldn't this be `True`? :thinking: That is at least what I was expecting the deletion-logic to ensure: that `/documents/originals` still is a directory, even if the sub-directories have been deleted. Maybe the deletion-logic is flawed and deletes one-too-many directories?

src/documents/models.py
BastianPoe: Fixed

src/documents/tests/test_file_handling.py
BastianPoe: Fixed

BastianPoe:
@pitkley Thanks again for your help - and again the PR got a lot better through your feedback. I have fixed the issues and this feel ready for merging now. What do you think?

src/documents/models.py
```diff
@@ -267,17 +277,91 @@ def __str__(self):
...
+
+        return mydictionary
+
+    @staticmethod
+    def fill_list(input_list, length, filler):
+        while len(input_list) < length:
+            input_list.append(slugify(filler))
+
+        return input_list
+
```
pitkley, line 318:
```suggestion
```

This method isn't used anymore, I think we should get rid of it.

BastianPoe: Removed

src/documents/models.py
```diff
@@ -314,6 +391,93 @@ def thumbnail_file(self):
...
+        # Try to delete the current directory
+        try:
+            os.rmdir(directory)
+        except os.error:
+            # Directory not empty, no need to go further up
+            return
+
+        # Cut off actual directory and go one level up
+        directory, _ = os.path.split(directory)
+        directory = os.path.normpath(directory)
```
MasterofJOKers, line 414: I think this code would really profit from using pathlib. E.g. we could iterate over  `https://docs.python.org/3/library/pathlib.html?highlight=pathlib#pathlib.PurePath.parents` instead of using `while` here.

But we don't use pathlib everywhere else, so probably not the right thing for this PR.

src/documents/models.py
```diff
@@ -314,6 +391,93 @@ def thumbnail_file(self):
...
+    # Determine the full "current" path
+    path_current = instance.filename_to_path(instance.filename)
+
+    # Move file
+    try:
+        os.rename(path_current, path_new)
+    except PermissionError:
+        # Do not update filename in object
+        return
+
```
MasterofJOKers, line 447: What if we get killed right here? `instance.filename` is still the old one, but the file is already at the new filename. Can we handle that somehow? Because in the next run, it will just fail in the `os.rename()` above, right?

In general, if something else is accessing the files, it might - by accident - move them around. I think we should not crash then. It raises a `FileNotFoundError`, which we need to handle some way here.

src/documents/tests/test_file_handling.py
```diff
@@ -0,0 +1,333 @@
...
+        # Check proper handling of files
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals/test"), True)
+        self.assertEqual(os.path.isdir(settings.MEDIA_ROOT +
+                         "/documents/originals/none"), True)
+
+        # Cleanup
+        os.remove(settings.MEDIA_ROOT +
+                  "/documents/originals/none/none-0000001.pdftest")
+        os.rmdir(settings.MEDIA_ROOT + "/documents/originals/none")
```
MasterofJOKers, line 144: Does this cleanup happen if we fail on the asserts? I'd prefer not having stuff hanging around, even though we use a new temporary directory for every test. I'm not using docker, so it would fill up my temporary directory.

src/documents/models.py
BastianPoe: Valid feedback! I have addressed this through change in source_filename. As we always append the document id to all filenames, we habe a fair change to find documents in case the rename failed.

What we cannot cover is someone just renaming on the filesystem level and then removing the document if suffix.

src/documents/tests/test_file_handling.py
BastianPoe: It did not - but now it does. I have added a tearDown to ensure, that all temp directories are deleted after running the tests.

src/documents/models.py
BastianPoe: I am happy to migrate to path lib, but it is not really used anywhere in paperless. What do you think about doing this change?

pitkley: While I agree that pathlib seems to be _the_ way to go when handling paths, I don't think we have to block this PR on that. Maybe it makes sense to refactor this separately?

MasterofJOKers: Yes, we should totally do it in probably multiple different PRs, so it's easier to work through.

MasterofJOKers:
@BastianPoe Could you please add a test for the source_filename thing you fixed? Since the tests worked before (I assume), we probably don't have a test for it, yet. If we want to refactor stuff (e.g. for pathlib), then we should make sure we don't optimize your fixes out of the code again.

BastianPoe:
I did add tests for this: test_document_renamed and test_document_renamed_encrypted

MasterofJOKers:
> I did add tests for this: test_document_renamed and test_document_renamed_encrypted

Sorry, I somehow missed them.

src/documents/tests/test_file_handling.py
```diff
@@ -0,0 +1,441 @@
...
+from dateutil import tz
+from django.test import TestCase, override_settings
+
+from django.utils.text import slugify
+from ..models import Tag, Document, Correspondent
+from django.conf import settings
+
+
+class TestDate(TestCase):
+    def tearDown(self):
```
MasterofJOKers, line 18: If we ever start running tests in parallel, which we currently don't seem to do, this will hurt us, as it will delete temporary directories of other tests, too. What do you think about making this forward-compatible and just remove `MEDIA_ROOT` by saving it on the instance? Something like `if hasattr(self, '_tmp_media_root'):` or something. Do you think that's overkill?

MasterofJOKers:
The new tests failed btw. https://travis-ci.org/github/the-paperless-project/paperless/jobs/673409471

MasterofJOKers:
I don't know why travis-ci isn't reporting in, but the tests are still failing https://travis-ci.org/github/the-paperless-project/paperless/jobs/674292249

BastianPoe:
@MasterofJOKers thank you very much for your review. Interestingly, the tests passed locally but not remotely. While investigating, I found some more bugs, added tests for them as well and fixed them of course. Please review and comment and/or approve. Thanks!
@pitkley please have another look and comment and/or approve. Thanks!

src/documents/tests/test_file_handling.py
BastianPoe: Done

src/documents/tests/test_file_handling.py
```diff
@@ -0,0 +1,616 @@
...
+from dateutil import tz
+from django.test import TestCase, override_settings
+
+from django.utils.text import slugify
+from ..models import Tag, Document, Correspondent
+from django.conf import settings
+
+
+class TestDate(TestCase):
+    deletion_list = []
```
pitkley, line 18: I feel like the `deletion_list` might be a bit fragile, at least if a new test is added and one forgets to call `add_to_deletion_list`.

One alternative I can think of is to have a media-root that is global to the test-case, e.g.:

```python
from tempfile import TemporaryDirectory

class TestDate(TestCase):
    media_root = TemporaryDirectory(prefix="paperless-test-")

    def tearDown(self):
        self.media_root.cleanup()

    # ...
    @override_settings(MEDIA_ROOT=os.join(media_root.name, "file-renaming")
    @override_settings(...)
    def test_file_renaming(self):
        # ...
```

I think that this would not only get rid of the deletion-list, making it easier to understand the existing tests and also easier to write new ones, it also helps in the tests being cross-platform. (I'm not sure if they currently are anyway, but we don't have to make the situation worse if we can help it.)

BastianPoe: This was indeed too fragile and I fixed this now. I found a way to handle this in setUp and hence this cannot be forgotten anymore.

BastianPoe:
@pitkley and @MasterofJOKers I have fixed the recent comments. Could you please have a look again? Thanks!

BastianPoe:
@MasterofJOKers, could you please have another go at the review? Thanks!

totti4ever:
Anything one can do to support this? I am eagerly waiting for this PR to finally use paperless! :-)

Issue #596, opened 2020-01-03
Preview to the side in edit mode


joschi36:
Mostly when i want to edit the document information it would be very useful to see the pdf to the side.

![image](https://user-images.githubusercontent.com/4538306/71726720-31657380-2e38-11ea-9cd7-4b7a1862f34d.png)

Maybe it would be a good idea to embed the PDF with `<object>`.

_PS.: A rotate option would be useful too as you can see ;)_

ddddavidmartin:
Yes please! I'd love to see this implemented. As a start I would even be happy to have a link to the document there.

bauerj:
@joschi36 @ddddavidmartin I have implemented something like this in #597.

It would be nice if one of you could test whether this works for them.

Nicras:
I would love to test this but I don't know how. I use docker

bauerj:
@Nicras If you use a recent version of paperless, this should be fairly easy:

    git remote add bauerj https://github.com/bauerj/paperless
    git fetch bauerj
    git checkout pdf-preview
    docker-compose build
    docker-compose up -d

Nicras:
I don't ;-)

I built my own docker containers without compose. But I managed to get the src folder persistent an changed the files accordingly and it works like a charm.

Makes editing documents a lot more efficient

<img width="1920" alt="2020-01-14_15-40-27" src="https://user-images.githubusercontent.com/35510003/72353599-9c3e6680-36e4-11ea-892a-ad53cdb4c5cc.png">

bauerj:
Thanks for testing! What theme are you using there?

Nicras:
It's just css modification to align with the style of my organizr (plex theme)
I did it by myself. I would love to share more screenshots but it's hard to grey out everything sensitive.
And I can't attach my overrides.css

Please let me know if you wanna try, I'll send it to you

PR #597, opened 2020-01-03
Add PDF preview next to edit form

bauerj:
This adds a simple PDF preview next to the edit form. Adding the CSS to the page footer is a bit ugly. If anyone with better knowledge of Django knows of a cleaner way to do this, please let me know.

On a large enough screen, it looks like this:
image

If there is not enough space, the preview will be shown below the form:
image

Closes: #596

### DIFF
--- a/src/documents/templates/admin/documents/document/change_form.html
+++ b/src/documents/templates/admin/documents/document/change_form.html
@@ -3,6 +3,10 @@
 {% block content %}
 
 {{ block.super }}
+<div class="side-preview">
+  <h2>Preview</h2>
+  <object data="/fetch/preview/{{object_id}}"></object>
+</div>
 
 {% if next_object %}
 	<script type="text/javascript">//<![CDATA[
@@ -16,6 +20,37 @@
 {% endif %}
 
 {% endblock content %}
+
+{% block extrastyle %}
+{{ block.super }}
+<style>
+.side-preview {
+    width: 100%;
+    height: 800px;
+    clear: both;
+}
+
+.side-preview object {
+    height: 100%;
+    width: 100%;
+}
+
+@media screen and (min-width: 1500px) {
+    #content-main {
+        width: 50%;
+    }
+    #footer {
+        padding: 0;
+    }
+    .side-preview {
+        float: right;
+        width: 40%;
+        height: 80vh;
+        clear: none;
+    }
+}
+</style>
+{% endblock %}
 
 {% block footer %}
 
--- a/src/documents/views.py
+++ b/src/documents/views.py
@@ -69,7 +69,10 @@ class FetchView(SessionOrBasicAuthMixin, DetailView):
             content_type=content_types[self.object.file_type]
         )
 
-        DISPOSITION = 'inline' if settings.INLINE_DOC else 'attachment'
+        DISPOSITION = (
+            'inline' if settings.INLINE_DOC or self.kwargs["kind"] == 'preview'
+            else 'attachment'
+        )
 
         response["Content-Disposition"] = '{}; filename="{}"'.format(
             DISPOSITION, self.object.file_name)
--- a/src/paperless/urls.py
+++ b/src/paperless/urls.py
@@ -37,7 +37,7 @@ urlpatterns = [
 
     # File downloads
     url(
-        r"^fetch/(?P<kind>doc|thumb)/(?P<pk>\d+)$",
+        r"^fetch/(?P<kind>doc|thumb|preview)/(?P<pk>\d+)$",
         FetchView.as_view(),
         name="fetch"
     ),


ahyear:
is it possible to add a setting to make it configurable:
"I want or I don't want a preview" (loading time issue for exemple...)

bauerj:
Yes, that is possible and sounds like a good addition. I would enable this feature by default though.

joschi36:
works like a charm thank you :+1:

ddddavidmartin:
This looks great! It works fine on my computer but it doesn't show the preview on mobile for me. I tried both Firefox and the stock web browser on my Android phone. Does anyone else have this issue?

bauerj:
Yes, I think this is on purpose. It would probably slow things down too much.

ddddavidmartin:
> Yes, I think this is on purpose. It would probably slow things down too much.

You mean it not showing on mobile is on purpose? It still shows the `Preview` heading for me, just not the actual document. But then the inline viewing of the documents doesn't work either. In the case of opening inline documents it just triggers a download of the PDF for me. I'm wondering whether it is more a case of the mobile browser not being able to show inline documents. 🤔

bauerj:
> I'm wondering whether it is more a case of the mobile browser not being able to show inline documents. 🤔

This rather seems to be the effect, not the cause.

ddddavidmartin:
You are a legend for implementing this @bauerj! This makes editing newly scanned documents so much easier.

I'm not too worried about editing documents on mobile, it just seemed odd that it doesn't work. Looks like it is a mobile browser limitation.

Happy to have this merged in.

sceiler:
@ddddavidmartin @bauerj works like a charm for me. It also shows up on my iPhone Xs, iOS 13. I think it depends on your phone and the browser used.

lbischof:
I often edit newly scanned documents on my smartphone. Instead of a preview I would prefer a download link. Maybe this could be configurable so it is possible to toggle between preview and download link/button?

bauerj:
Which browser do you use on your phone? For me it already does just that.

lbischof:
@bauerj Are you replying to my question? Do you mean that the `<object>` is rendered as a link in some browsers? 

Current situation: after consumption I send a link to my phone. This link brings me directly to the edit page (so I can add tags, check date etc) and as far as I can see I cannot currently download the document from the "edit" page.

bauerj:
> Do you mean that the <object> is rendered as a button in some browsers?

Yes. As far as I can tell in mobile browsers.

src/documents/templates/admin/documents/document/change_form.html
```diff
@@ -26,4 +30,37 @@
 		django.jQuery(".field-created input").first().attr("type", "date")
 	</script>
 
+	{# Hack to make preview appear next to form #}
```
pitkley: The following should work and be the intended way:

```jinja
{% block extrastyle %}
{{ block.super }}
<style> ... </style>
{% endblock %}
```

Make sure to add this outside of the `block footer` section. 🙂

bauerj: Thanks, I've changed the template accordingly.

bauerj:
Any chance to get this reviewed by someone?

bauerj:
Thanks @ddddavidmartin!

ddddavidmartin:
Awesome work on this one @bauerj! I have been running this locally for weeks now and I find it hard to believe how we managed without it.

> Any chance to get this reviewed by someone?

Apologies for the long wait on this one! I completely missed that I had to review it again after your change. I just tried the latest build and it looks fine to me.

stueja:
Implemented the changes, see the "Preview" tag on the right side, but no preview, neither in Firefox nor in Chromium.
HTML source looks correct so far.
```
<div class="side-preview">
  <h2>Preview</h2>
  <object data="/fetch/preview/967"></object>
</div>
```
I am not on the dockerized version though. Does that make a difference?

sceiler:
@stueja Where or how you run this should not affect the preview. This is most likely a client issue. For example I am using Windows 10 and Firefox 72 and Chrome 80 shows the preview. I also tested it on Safari 10.14 on an MacBook and Safari on an iPhone XS 13.
It might be related to your installed plugins or the security settings of your browser? Does the preview work on your phone?

stueja:
Hey @sceiler thanks for your answer! I have tried on 
 - Arch Linux, Firefox 74.0, with several plugins
 - Arch Linux, Chromium 80.0.3987.149, without any plugins
 - iPhone XS 13.3, Safari
 - (i can try later on Windows 10, Firefox, Chrome and Opera)
and I am sorry, I can't see a preview.
I added the changes to release 2.7.0. Am I missing prerequisites from other merges?

bauerj:
What do you see when you open /fetch/preview/967 in your browser?

stueja:
@bauerj, I see the pdf.

bauerj:
> I see the pdf.

Okay, that's a good start 😃 

How exactly did you add these specific changes to 2.7.0? Do you see any error messages in your browser console? What do you see in stead of the preview? An error message?

stueja:
I added the changes by opening the given file in vim, searching for the position, and copy-pasting them. Instead of a preview I see -- nothing. No error message in the page or html source.

However, the idea of the browser console shows 
```
Load denied by X-Frame-Options: “DENY” from “https://paperless.js73.de/fetch/preview/967”, site does not permit any framing. Attempted to load into “https://paperless.js73.de/admin/documents/document/967/change/”.
```
After a short googling, Stackoverflow showed me how to hide the X-Frame-Options header
(https://stackoverflow.com/questions/47405597/x-frame-options-in-nginx-to-allow-all-domains)
```
proxy_hide_header X-Frame-Options;
```
and voilá, I can see the preview! <3 Thanks alot for your support!

PR #599, opened 2020-01-06
Allow non-unique GIDs

ljm42:
Need to allow non-unique group ids, so paperless can map to an existing GID.

The comment says the map_uidgid() function came from docker-gitlab, the same fix was applied there too:
  https://github.com/sameersbn/docker-gitlab/commit/c80bd876628d0c04f8ef2a3ffb46fe899bf63d23


ljm42:
More info:

Inside the paperless docker, files created by paperless in the media folder are owned by `paperless:paperless`, that is fine.

Outside the docker, I want them to be owned by `nobody:users`

To do this, I need to pass these variables to the docker:
```
  -e 'USERMAP_UID'='99'
  -e 'USERMAP_GID'='100'
```

If I just send the USERMAP_UID variable it changes the UID without a problem, but files are owned by `nobody:1000`

When I send the USERMAP_GID variable, the docker fails with this message:
```
Mapping UID and GID for paperless:paperless to 99:100
groupmod: GID '100' already exists
```

This PR will allow the groupmod command to succeed


As mentioned above, this same fix was made in the original repo that the map_uidgid() function came from:
  https://github.com/sameersbn/docker-gitlab/commit/c80bd876628d0c04f8ef2a3ffb46fe899bf63d23

jholton:
Running into this problem right now.   Hopefully gets merged soon!

PR #610, opened 2020-02-09
Add user and group paperless and avoid to use hardcoded userId and groupId

michalharakal:
1000 hardwired its not working e.g. on synology.

Related to #306

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,6 @@ LABEL maintainer="The Paperless Project https://github.com/the-paperless-project
 
 # Copy Pipfiles file, init script and gunicorn.conf
 COPY Pipfile* /usr/src/paperless/
-RUN addgroup -S paperless && adduser -S paperless -G paperless
 COPY scripts/docker-entrypoint.sh /sbin/docker-entrypoint.sh
 COPY scripts/gunicorn.conf /usr/src/paperless/
 
@@ -49,6 +48,7 @@ RUN apk add --no-cache \
 # Create the consumption directory
     mkdir -p $PAPERLESS_CONSUMPTION_DIR && \
 # Create user
+    addgroup -S paperless && adduser -S paperless -G paperless && \
     chown -Rh paperless:paperless /usr/src/paperless && \
     mkdir -p $PAPERLESS_EXPORT_DIR && \
 # Setup entrypoint
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ LABEL maintainer="The Paperless Project https://github.com/the-paperless-project
 
 # Copy Pipfiles file, init script and gunicorn.conf
 COPY Pipfile* /usr/src/paperless/
+RUN addgroup -S paperless && adduser -S paperless -G paperless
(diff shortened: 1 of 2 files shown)


Tooa:
@michalharakal Is this related to https://github.com/the-paperless-project/paperless/pull/599?

Dockerfile
```diff
@@ -48,8 +49,6 @@ RUN apk add --no-cache \
 # Create the consumption directory
     mkdir -p $PAPERLESS_CONSUMPTION_DIR && \
 # Create user
-    addgroup -g 1000 paperless && \
```
bauerj, line 51: If you leave the comment here, you should move the command back down.

bauerj:
What exactly is the advantage of this? Wouldn't `adduser -S paperless -G paperless` simply use the first unused UID for the new user? Since there are no other users in that Docker image that would be 1000, wouldn't it?

Dockerfile
michalharakal: Yes, it makes sense. Should be on the right place now.

michalharakal:
I can't explain why 1000 is not working on synology. As I understood, its generally a good practice even in container not to run as root...

michalharakal:
@Tooa yes indeed, #599 seems to be a different solution to the same problem.

bauerj:
> As I understood, its generally a good practice even in container not to run as root.

Surely, but UID 1000 is not root (that would be UID 0). Of course I'm not opposed to any other number I just want to understand the issue in more detail.

michalharakal:
@bauerj Meanwhile I have checked #599 and paperless starts on my synolgy out of the box, so its working for me now and my merge ist not required anymore.  I am no expert neither in linux nor in docker, but from by experience it feels somehow not realy right to depend on hard wired constants. What I already saw it that user/group and rights are handled in entrypoint script dynamically. There is a reason for that I don't know (probably issues with mounting volumens etc.)
```
#!/bin/bash

if [ -z "${USER_UID}" -o -z "${USER_GID}" ]; then
    echo "Getting UID and GID from current directory"
    CURRENT_DIR=$(ls -lnd .)
    USER_UID=$(echo "${CURRENT_DIR}" | cut -f3 -d' ')
    USER_GID=$(echo "${CURRENT_DIR}" | cut -f4 -d' ')
fi

if [ "${USER_UID}" -ne 0 ]; then
    groupadd \
        --gid ${USER_GID} \
        paperless
    
    useradd \
        --uid ${USER_UID} \
        --gid ${USER_GID} \
        --shell /bin/bash \
        paperless
fi

echo "Using UID ${USER_UID} and GID ${USER_GID} for execution"

chown --quiet --recursive "${USER_UID}:${USER_GID}" /home/paperless

echo "Executing $@"

if [ "${USER_UID}" -ne 0 ]; then
    sudo --user paperless --preserve-env --set-home -- "$@"
else
    exec "$@"
fi

```

michalharakal:
Anyway, I can close the MR.

bauerj:
I'm glad it works for you now! I agree that it's not a good idea to use some "magic" constants like 1000. In this case, it might be better to make that configurable.

PR #601, opened 2020-01-11
Fix warnings in bash code blocks

Tooa:
### Summary 

* Fixes warning generating docs: `WARNING: Could not lex literal_block as "bash". Highlighting skipped`.

### Addressed Issues

* See https://github.com/sphinx-doc/sphinx/issues/4098

### Expected Behaviour 

* `$ sphinx-build -b html ../docs ../docs/_build -W` builds documentation successfully

### Actual Behaviour 

* Building documentation fails with error
* This issue also occurs in the Travis build pipeline

```bash
(paperless) user:paperless/src$ sphinx-build -b html ../docs ../docs/_build -W
Running Sphinx v2.3.1
loading intersphinx inventory from http://docs.python.org/objects.inv...
intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 15 source files that are out of date
updating environment: [new config] 15 added, 0 changed, 0 removed
reading sources... [100%] utilities                                                                                       
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] utilities                                                                                        

Warning, treated as error:
/paperless/docs/utilities.rst:242:Could not lex literal_block as "bash". Highlighting skipped.
```

### Conducted Tests after Patch

```bash
$ sphinx-build -b html ../docs ../docs/_build -W
$ firefox docs/_build/index.html
```
paperless-docs-utilities

:arrow_right: still looks good with patch

PR #612, opened 2020-02-16
Upgrade Docker image to Alpine 3.11

languitar:
This make tesseract 4.1 avaialbe, which fixes some things like infinite
processing loops on some documents:
https://github.com/tesseract-ocr/tesseract/issues/2288#issuecomment-561727361

Some dependencies had to be bumped for being compatible with the new Alpine
libraries.

### DIFF
--- a/src/paperless/settings.py
+++ b/src/paperless/settings.py
@@ -60,7 +60,7 @@ if _allowed_hosts:
     ALLOWED_HOSTS = _allowed_hosts.split(",")
 
 FORCE_SCRIPT_NAME = os.getenv("PAPERLESS_FORCE_SCRIPT_NAME")
-    
+
 # Application definition
 
 INSTALLED_APPS = [
@@ -109,7 +109,7 @@ MIDDLEWARE = [
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
 
 # We allow CORS from localhost:8080
-CORS_ORIGIN_WHITELIST = tuple(os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "localhost:8080").split(","))
+CORS_ORIGIN_WHITELIST = tuple(os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "http://localhost:8080,https://localhost:8080").split(","))
 
 # If auth is disabled, we just use our "bypass" authentication middleware
 if bool(os.getenv("PAPERLESS_DISABLE_LOGIN", "false").lower() in ("yes", "y", "1", "t", "true")):
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.10
+FROM alpine:3.11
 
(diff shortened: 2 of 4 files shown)


languitar:
Travis failure seem to be unrelated to my changes. They originate from the sphinx documentation :/

src/paperless/settings.py
```diff
@@ -109,7 +109,7 @@ def __get_boolean(key, default="NO"):
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
 
 # We allow CORS from localhost:8080
-CORS_ORIGIN_WHITELIST = tuple(os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "localhost:8080").split(","))
+CORS_ORIGIN_WHITELIST = tuple(os.getenv("PAPERLESS_CORS_ALLOWED_HOSTS", "http://localhost:8080").split(","))
```
bauerj: How is this related to upgrading the operating system?

languitar: Sorry, forgot to mention this in the commit message. After upgrading the packages in the lock file, some upgraded dependency for cors now requires full URIs instead of hostnames now.

bauerj: Okay, thanks for clarifying :+1:

Tooa:
> Travis failure seem to be unrelated to my changes. They originate from the sphinx documentation :/

@languitar @bauerj Merge PR https://github.com/the-paperless-project/paperless/pull/601 first to fix this issue

pitkley:
@languitar thanks for the PR, dependency updates are always nice! #601 is just missing another review from @the-paperless-project/reviewers, then you can rebase on master to fix the Travis issue.

Regarding the `setrlimit(RLIMIT_CORE)` error/warning @Tooa mentioned: a simple fix is to `echo 'Set disable_coredump false' >> /etc/sudo.conf`, see https://ask.fedoraproject.org/t/sudo-setrlimit-rlimit-core-operation-not-permitted/4223 and https://gitlab.alpinelinux.org/alpine/aports/issues/11122. Maybe the suggested `doas`-change in the latter link is a more correct fix, but I think that would be too big for what this PR should be doing.

Feel free to ping me directly for a review once the Travis build is fixed. 👍

languitar:
I have rebased the PR, added the setrlimit fix, and split the PR into two distinct commits with further clarifications.

src/paperless/settings.py
MasterofJOKers: Can we mention this setting in [the docs](https://github.com/the-paperless-project/paperless/blame/master/docs/setup.rst#L196)? I would expect people having SSL enabled to change this. Or can we get away with `//` instead of `http://` to support both?

languitar: I think a proper URL is required now. So // probably doesn't work.

bauerj: You could just add both URLs, couldn't you?

languitar: done

Issue #635, opened 2020-04-27
update gunicorn requirement to 20.0.4


apiontek:
Running paperless on python 3.8 (on ubuntu 20.04) with gunicorn 19.9.0 yields this error:

```
/usr/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
```

This was resolved in gunicorn 20.0 per [this issue](https://github.com/benoitc/gunicorn/issues/2091) and doing a `pip install -U gunicorn` in my paperless venv gets rid of the error message for me.

Tooa:
Hi, thanks for reporting this issue with a working solution. Are you interested in submitting a PR to fix this issue?

apiontek:
Done

PR #636, opened 2020-04-28
bumping gunicorn version requirement to 20.0.4

apiontek:
updated gunicorn in requirements.txt to address #635

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ faker==2.0.0
 filelock==3.0.12
 filemagic==1.6
 fuzzywuzzy[speedup]==0.15.0
-gunicorn==19.9.0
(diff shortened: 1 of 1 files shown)


diveflo:
We should probably update it in for pipenv in the Pipfile and Pipfile.lock as well, right? Otherwise, the docker image won't be using the same version as manual installs

pitkley:
Judging from [this changelog-entry](https://github.com/the-paperless-project/paperless/blame/master/docs/changelog.rst#L222-L223), `Pipfile`/pipenv was added but the full migration to it never happened. I am very strongly in favour of phasing out `requirements.txt` in favour of `Pipfile.lock`; I opened #661 to keep track of this.

Issue #613, opened 2020-02-18
Paperless web UI responds slowly on Docker


lawtancool:
I'm running Paperless using the example `docker-compose.yml` provided. When trying to browse documents, it can take over 15 seconds for the page to load. 
I connect to Paperless directly over my local network through a port exposed in Docker, my database is stored on an SSD, and my system is not overloaded from a CPU/RAM perspective.

lawtancool:
I was able to speed up the web UI significantly by editing this line in `scripts/gunicorn.conf`:
Old: `workers = 1`
New: `workers = 9`

http://docs.gunicorn.org/en/latest/design.html#how-many-workers seems to suggest that 1 worker simply wasn't enough for proper operation.

PR #641, opened 2020-05-04
Increase gunicorn workers from 1 to 4

lawtancool:
Having only 1 gunicorn worker results in extremely slow web UI load speed, raising it to 4 should improve performance significantly without overwhelming low end server environments. 

See [https://docs.gunicorn.org/en/stable/design.html#how-many-workers](https://docs.gunicorn.org/en/stable/design.html#how-many-workers) for more details

This resolves issue #613.

### DIFF
--- a/scripts/gunicorn.conf
+++ b/scripts/gunicorn.conf
@@ -1,6 +1,6 @@
 bind = '127.0.0.1:8000'
 backlog = 2048
-workers = 4
+workers = 3
 worker_class = 'sync'
 worker_connections = 1000
 timeout = 20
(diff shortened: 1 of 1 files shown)


MasterofJOKers:
Reading the docs you provided, the minum recommended setting would be `2 = 1 core + 1`.

How about setting it to `2` and document setting it to higher values via running the container with `gunicorn --workers 4`. Or does setting it on the commandline not work if it's already set in the config?

lawtancool:
@MasterofJOKers The documentation actually recommends `(2 x $num_cores) + 1` workers, so assuming a single core system that would be `(2 x 1) + 1 = 3 workers. I've added a commit to change to this number.

PR #642, opened 2020-05-04
Add link to Paperless App to Related Projects

bauerj:
Hey,

I'm developing a native Android/iOS frontend for Paperless using Flutter at https://github.com/bauerj/paperless_app.

I didn't have much users testing it yet so I would be glad if anyone could give any feedback. It has limited functionality (filter, search, open PDF) but I plan to extend it with more features like an offline mode.

@danielquinn Do you think it would be a good idea to move this project to the @the-paperless-project organisation? I think it could improve visibility and open source contributions.


bauerj:
@MasterofJOKers @pitkley Sorry to bother you but after reading #470 it seems like Daniel Quinn is busy enough with other stuff. 

Could you please look into this and let me know if you would able to merge this (and maybe move the repository to the @the-paperless-project organisation)?

bauerj:
Sure, I didn't know the two-reviewer-rule would apply to that repository too then. In that case I agree it wouldn't be a good idea. Of course I want to keep maintaining the app.

PR #670, opened 2020-06-02
Fix warning in models.py

frrad:
When I ran `docker-compose run --rm webserver createsuperuser` I got the following warning:

```
/usr/src/paperless/src/documents/models.py:339: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if delimiter is -1:
/usr/src/paperless/src/documents/models.py:342: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if delimiter is -1:
```

### DIFF
--- a/src/documents/models.py
+++ b/src/documents/models.py
@@ -336,10 +336,10 @@ class Document(models.Model):
             # Find delimiter
             delimiter = t.name.find('_')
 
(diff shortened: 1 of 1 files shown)


frrad:
I think this fixes https://github.com/the-paperless-project/paperless/issues/649 fwiw

Issue #337, opened 2018-04-08
Does the Docker image support Raspberry Pi (ARM)?


crouthamela:
I'm looking to add this to my RPi server. Will the Docker image work on this? Are there any processes that may fail?

addadi:
The docker image is based on alpine 3.7 which is [x86_64](https://hub.docker.com/r/_/alpine/). I guess that if you want to support RPi you would need to use arm image (first line in the dockerfile)

danielquinn:
Hmm.  That's a good point.  It'd be nice to be able to have different images for different architectures.  @addadi, is that easy to do in Dockerland?  Would you normally do it as a separate repo, or different directory, or...?

crouthamela:
It shouldn't be too hard to add an official dockerfile. For example [netdata](https://hub.docker.com/r/firehol/netdata/) does this. You just pull the respective tag and it uses the appropriate dockerfile.

danielquinn:
Ah, interesting.  They just suffix the file with the architecture: `Dockerfile.armv7hf` for example.  @crouthamela if you wanna send me a pull request with this sort of thing working for your Pi, I'll merge it and add it to the official Hub page.

stgarf:
@crouthamela Did you ever ending up building an image for your RPi? If you built a working container, submit a PR with your `Dockerfile` here and it can probably be wired up with Travis-CI and DockerHub.

crouthamela:
No, sorry. I ended up upgrading to a NUC.

fipstree:
Are there any news on this topic? Does a docker file exist for raspberry pi?

diveflo:
Have you tried it?
I have it running on my raspberrypi 4 and a RockPi4 - for the latter, my only issue was getting docker and docker-compose compiled for arm64v8. On the raspberrypi 4 running raspian buster it just worked out of the box.

maroessler:
Are there any plans to add an official paperless docker image for arm32v7? I have made my own Dockerfile which basically just changes the first line to use the arm32v7 version of alpine (see my [fork](https://github.com/maroessler/paperless)). I used this image successfully on my Pi 3B+.

diveflo:
We could use buildx to create multi-platform images. The build seems to work fine locally via 
`docker buildx build -t floriang89/paperless-arm:latest --platform linux/amd64,linux/arm64,linux/arm/v7
.`
[...]

PR #674, opened 2020-06-08
Build and publish Multi-Arch Docker Images

MarkSchmitt:
First things first: I love paperless, it's awesome!

In order to use it on multiarch (amd64, arm64, arm32v7) kubernetes cluster, I've created docker images for all archs that I think are useful and added a manifest file, so  docker/containerd can automatically pick and choose the correct operating system/architecture.

It'll detect if it's on a tag and if so, tag the docker image like so.

You'll need to set the following travis environment varibles:
- BUILD_DOCKER = 1
- DOCKER_USERNAME
- DOCKER_PASSWORD
- DOCKER_REPO

As far as I can tell, all but the DOCKER_REPO are already set, that's what used to be the "tag" in the existing CI script. I wanted to have that flexible, so I can set it to my tag in my pipeline instead.

I've extended the travis-ci pipeline to use arm64 for the arm64 build, also see: https://docs.travis-ci.com/user/multi-cpu-architectures#using-docker-in-multiple-cpu-architecture-based-builds-within-lxd-containers

For arm32v7 and arm32v6 there is afaik currently no native support, so instead, I'm using qemu userland emulation to build it, executing on the standard amd64 hardware. See also: https://github.com/multiarch/qemu-user-static#getting-started
That has one drawback: Because docker is not aware that the code executed in the docker container is arm32v7/v6, it'll still pull the amd64 base image (btw, a multi arch image too). That is why there's a sed-replace, setting the base image to a specific version that I know is arm32v7/v6.

Same can be done for any arch that is supported in qemu, has a base image we can use and is supported in paperless and all it's dependencies.

Hope you like the pull request. Let me know if you need anything changed to accept it. I'd like to get my things back in upstream and stop maintaining my own branch :).


MarkSchmitt:
A couple more things: the build in travis is failing because of the missing DOCKER_REPO environment variable, setting it to "the-paperless-project/paperless" should do the trick.

That issue should get resolved if we can get this Pull Request merged: #337 

Docker buildx seems to be discussed as solution to build cross architecture sometimes, also in that issue #337 . I have not had any luck getting that to work, not in a pipeline, not locally on a linux box and I've never seen a build pipeline using it.

whyicantusemyemailasusername:
Are there any eta for this? I'm waiting to be able to use an official image on RPI4
Thanks!

.travis.yml
```diff
@@ -7,26 +7,126 @@ before_install:
...
+            - true
+          after_success:
+            - true
+          script:
+            - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+            # register binfmt stuff for qemu-static binaries so we can use userland-emulation
+            - docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+            # replace the multi-arch reference with a specific, arm32v7 version. else docker will use the platform specific one,
+            # which is amd64.
+            - sed -i 's/FROM alpine:3.10/FROM alpine@sha256:2632d6288d34d7175021683f6e363fa7c0fa8866a565eb285e36e3b856545e82/g' Dockerfile
```
sbrunner, line 64: I'm not sure that using the sha is good for maintenance...

MarkSchmitt: indeed using the sha directly was not my preferred solution. the underlying problem I'm trying to solve is, that the alpine:3.10 image is a multi-arch image. when using qemu in docker to emulate the arm32v7 cpu architecture on x86_64, docker still pulls the x86_64 image instead of the arm32v7 one. by specifying the image directly and not the manifest (which points to the individual images depending on the operating system and architecture) we can circumvent that problem.

I tried telling docker which architecture I would like it to build for, but failed finding any usable documentation and only vague hints towards buildx, which seems to be a complete mess :shrug: 

So I'm open to suggestions - but I don't see how to solve this without tagging a specific image.

sbrunner: Ok, thanks for your explain...

MarkSchmitt:
> From where does come from ${DOCKER_REPO}?
that's a travis-ci environment variable you need to set. it's not part of the pipeline, that way I can push into my repo in my travis build. we can go back to hardcoding it, though I don't see the advantage (forks will break in either case and making it harder to maintain forks)

> We need images without ${TRAVIS_COMMIT}.
that's always the git revision, so imho this is always set. we could not push non-tagged, non-master (latest) builds though.

> We need to have images with ${TRAVIS_TAG}.
that's supported, those lines are responsible: https://github.com/the-paperless-project/paperless/pull/674/files#diff-354f30a63fb0907d4ad57269548329e3R118-R127

> We need to have the image with tag latest for backward compatibility.
that's supported, those lines are reponsible: https://github.com/the-paperless-project/paperless/pull/674/files#diff-354f30a63fb0907d4ad57269548329e3R111-R116

> If you don't use the file ci/deploy-docker he should be removed but I think that he should be adapted to don't have many copy past code.
I'll remove that file in a follow up commit.

MarkSchmitt:
@sbrunner

Here you can find a latest image generated with the changed travis-ci pipeline:
https://hub.docker.com/layers/moztr/paperless-travis/latest/images/sha256-dec1c803503a52d4f1c6a981453e4c5c95dcc8aa2d523afe2981da6ef5bbb7d2?context=explore

And here one for a tag (I created for testing purposes)
https://hub.docker.com/layers/moztr/paperless-travis/0.0.2/images/sha256-61c8449316fa63b74536f75c47080cbe4f63d33af3fb07695f53f4dfaf83a306?context=explore

Anything else I can do to help you accept the pull request?

sbrunner (approved):
Looks good :-)

MarkSchmitt:
Moving to jonaswinkler/paperless-ng. Looks like this project is dead.

PR #677, opened 2020-06-13
Preserve file metadata when exporting

Tooa:
* Use [`copy2`](https://docs.python.org/2/library/shutil.html#shutil.copy2) instead of `copy` when exporting documents. This preserves documents metadata

* Useful when the export dir is used to `rsync` files to a remote host for offsite backups. The function `copy2` makes sure the modification date does not change when exporting on a regular basis. Otherwise, `rsync` copies all files to the target even if they already exist on the remote. One can work around this behavior with `--ignore-existing`,  but I think copying metadata backups with the export functionality is important.

### DIFF
--- a/src/documents/management/commands/document_exporter.py
+++ b/src/documents/management/commands/document_exporter.py
@@ -87,8 +87,8 @@ class Command(Renderable, BaseCommand):
 
             else:
 
-                shutil.copy(document.source_path, file_target)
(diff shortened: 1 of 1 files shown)


MasterofJOKers (approved):
Sounds like a good idea, even though probably more expensive on slow devices, because of the additional lookups and writes.

I'm in favor of doing this for the importer, too, even though the only use-case would be re-exporting again, I guess.

PR #702, opened 2020-10-17
Upgrade to Django 2.2 LTS

jayme-github:
Update paperless to the latest Django 2.2 (LTS) release. Django 2.0 is already EOL since April 2019.

I also dropped python 3.5 from the matrix (EOL September 2020) as it requires zipp 1.x and therefore extra rules in Pipenv to generate a working requirements.txt. If that's a problem let me know and I'll look into that.

Supersedes / Closes  #671


Sblop (approved):

Sblop (approved):

Sblop (approved):

Sblop (approved):

src/paperless_text/parsers.py
```diff
@@ -53,8 +53,14 @@ def create_bg():
...
         def read_text():
             with open(self.document_path, 'r') as src:
-                lines = [l.strip() for l in src.readlines()]
-                text = "\n".join([l for l in lines[:n_lines]])
+                lines = []
+                count = 0
+                for line in src:
+                    if count >= n_lines:
+                        break
+                    count += 1
```
MasterofJOKers, line 61: Imho better with [enumerate](https://docs.python.org/3/library/functions.html#enumerate)
```python
for count, line in enumerate(src, start=1):
  if count >= n_lines:
    break
```

MasterofJOKers (approved):
> and don't read the whole file if there is no need to

Should have been its own commit instead of hiding it in the second line of a commit message.

PR #708, opened 2020-11-03
Update scanners list to include mobile options

hannahswain:
Added a section for mobile phones and added new columns that are relevant to these apps. Added my suggestion.

### DIFF
--- a/docs/scanners.rst
+++ b/docs/scanners.rst
@@ -8,6 +8,9 @@ scanner you use, but sometimes finding a scanner that will write to an FTP,
 NFS, or SMB server can be difficult.  This page is here to help you find one
 that works right for you based on recommentations from other Paperless users.
 
+Physical scanners
+-----------------
+
 +---------+----------------+-----+-----+-----+----------------+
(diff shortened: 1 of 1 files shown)


danielquinn:
Hi Hannah!  Thanks for your contribution :-)  I assume you're recommending this scanner because you've used it yourself with Paperless?  It's funny, I never considered phone-based scanners when this project started.

danielquinn (approved):
👍🏻

hannahswain:
Yes, this is what I've been using rather than pulling out a scanner whenever I need to add documents. It's great because it'll also let you set up naming conventions, so I have a fill in field for correspondent, title, and tags. It'll then correctly name the document for Paperless to guess the correct values. I save it to a local folder on Android. SyncThing picks it up and moves it to the consume folder for me. So easy and then I don't end up with a backlog of stuff to scan.

kohfuchs (approved):
doing the same here, good idea to add it to the project.

kohfuchs (approved):

Sblop (approved):

MasterofJOKers (approved):

PR #716, opened 2021-01-14
Disable DjangoQL 0.14 advanced search syntax by default.

grembo:
DjangoQL 0.14 enables advanced search/completion by default. This patch
disables it again, so paperless behaves the same regardless of the
version used.

See
- https://github.com/ivelum/djangoql/blob/master/CHANGES.rst#0140
- https://github.com/ivelum/djangoql#using-djangoql-with-the-standard-django-admin-search

When using DjangoQL 0.13, this is simply a NOOP.

### DIFF
--- a/src/documents/admin.py
+++ b/src/documents/admin.py
@@ -196,6 +196,8 @@ class DocumentAdmin(DjangoQLSearchMixin, CommonAdmin):
 
     date_hierarchy = "created"
 
+    djangoql_completion_enabled_by_default = False
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.document_queue = []


MasterofJOKers:
It seems like we're still on 0.12.3. Is it a NOOP there, too?

Would it make sense to expose this as a setting, so the user can choose whether they want djangoql by default or not?

grembo:
> It seems like we're still on 0.12.3. Is it a NOOP there, too?

Yes, it only sets a local variable, which in case of older versions is simply ignored, as it's unknown to those.

> 
> Would it make sense to expose this as a setting, so the user can choose whether they want djangoql by default or not?

Maybe once you upgraded your preferred version to 0.14 (I'm maintaining the FreeBSD port, which uses latest versions by default, that's why I stumbled over this problem)? AFAIK the tunable isn't available in earlier versions and would have to be emulated by setting it explicitly for those, probably not worth the effort.

MasterofJOKers (approved):

PR #718, opened 2021-01-22
Typo

glogiotatidis:
Thanks for the great software!


MasterofJOKers:
Thank you for trying to contribute to it, but this seems to bo a duplicate of #717

glogiotatidis:
Indeed, didn't check for another open PR.

